PR #12, opened 2019-10-04
Redux and react navigation

maricalucian:


krisboit:
I will merge this, but some changes need to be done.

PR #24, opened 2019-10-15
Add dashboard screen

maricalucian:
This is work in progress.


src/components/account-card/account-card.tsx
```diff
@@ -0,0 +1,41 @@
...
+                    </TextSmall>
+                </Text>
+                <TextSmall>{props.account.address} </TextSmall>
+            </View>
+            <Text>></Text>
+        </View>
+    );
+};
+
+export default AccountCard;
```
krisboit: I'm not a fan of default exports, we should agree on a format and use it for all components. With named exports autocomplete is working better...

src/components/conversion-card/conversion-card.tsx
```diff
@@ -0,0 +1,41 @@
...
+                1
+            </Convert>
+            <TextSmall style={[change >= 0 ? styles.changeUp : styles.changeDown]}>
+                {change}
+            </TextSmall>
+        </View>
+    );
+};
+
+export default connect(
```
krisboit: As i said, i'm not a fan of default exports :), I'm thinking to do something like:
```
const ConversionCardComponent = () => {....}

export ConversionCard = connect(...)(ConversionCardComponent);
```
What do you think ?

src/components/account-card/account-card.tsx
maricalucian: I`m all in for consistency. Default exporting is just a habit for components, we can use named exports.

src/components/conversion-card/conversion-card.tsx
maricalucian: Yup, thats good going with named exports.

PR #31, opened 2019-10-21
update build number

georgeburduhos:


krisboit:
we will keep commit numbers as build number, as it's better for the moment

PR #34, opened 2019-10-22
transparent blockchain selector bar, revisit colors

maricalucian:


src/screens/dashboard/dashboard.tsx
```diff
@@ -163,26 +164,33 @@ export class DashboardScreenComponent extends React.Component<IProps & IReduxPro
...
-                                this.state.coinIndex === i && styles.blockchainButtonActive
-                            ]}
-                            onPress={() => this.setActiveCoin(i)}
-                        >
-                            <Text
-                                style={
-                                    this.state.coinIndex === i && styles.blockchainButtonTextActive
-                                }
+                <LinearGradient
+                    colors={['#00000000', '#000000CC']}
```
krisboit: use theme to set the colors. You have 2 options:
1. use styles, if it's possible
2. use useContext hook

krisboit: or we could pass the `theme` object to components from `withTheme`, this way we have direct access to theme props

maricalucian: updated.

PR #38, opened 2019-10-23
extension build setup

krisboit:


extension/background/background.ts
```diff
@@ -0,0 +1,2 @@
+// tslint:disable-next-line:no-console
+export default console.log('Hello world!!!');
```
georgeburduhos: Console log?

krisboit: :)) for now, it's just a test to see that the background script is loaded and working.

PR #43, opened 2019-10-24
Settings

georgeburduhos:


src/screens/settings/__tests__/settings.test.tsx
```diff
@@ -1,12 +1,64 @@
...
 describe('settings screen component', () => {
+    const wrapper = shallow(<SettingsScreenComponent {...props} />);
+
     it('renders correctly', () => {
-        const wrapper = shallow(<SettingsScreen />);
         expect(wrapper.debug()).toMatchSnapshot();
     });
+    it('Pin login switch tapped', () => {
+        wrapper.find('[testID="pin-login"]').simulate('valueChange');
+        expect(props.mock).toHaveBeenCalled();
```
krisboit: Since prop is a global value, i think that before each test we should clear mock, i think toHaveBeenCalled() will pass in all cases if one call was made.
https://jestjs.io/docs/en/mock-function-api.html#mockfnmockclear

georgeburduhos: Corect

PR #120, opened 2019-11-06
save password in keychain and encryped wallet info

maricalucian:


src/core/secure/storage.ts
```diff
@@ -0,0 +1,27 @@
...
+};
+
+export const readEncrypted = async (key: string, hash: string) => {
+    try {
+        const encryptedData = await AsyncStorage.getItem(`${KEY_PREFIX}${key}`);
+        if (encryptedData !== null) {
+            const data = await decrypt(encryptedData, hash);
+            return data;
+        }
+        return 'xxxxx';
```
krisboit: you should return here a Promise.reject();

src/core/secure/storage.ts
```diff
@@ -0,0 +1,27 @@
...
+    } catch (e) {
+        // error reading value
```
krisboit: same here, in case of an error the function is not returning anything

src/screens/create-wallet-confirm-mnemonic/create-wallet-confirm-mnemonic.tsx
```diff
@@ -160,3 +159,5 @@ export const CreateWalletConfirmMnemonicScreen = smartConnect(
         withTheme(stylesProvider)
     ]
 );
+
+CreateWalletConfirmMnemonicScreenComponent.navigationOptions = navigationOptions;
```
krisboit: i made some changes, i would say fixes in PR #118, so now static properties are moved on the enhanced component by `withTheme` enhancer. add `navigationOptions` on `CreateWalletConfirmMnemonicScreenComponent` before calling `smartConnect`

src/core/secure/storage.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const KEY_PREFIX = 'mw-';
+
+export const storeEncrypted = async (data: string, key: string, hash: string) => {
+    const encryptedData = await encrypt(data, hash);
+
+    try {
+        await AsyncStorage.setItem(`${KEY_PREFIX}${key}`, encryptedData);
+    } catch (e) {
+        // saving error
```
krisboit: return a `Promise.reject()` here, if `AsyncStorage.setItem()` is failing the function is returning no result.
Other approach would be to loose the `try...catch` and use it where `storeEncrypted()` is called.

src/screens/create-wallet-confirm-mnemonic/create-wallet-confirm-mnemonic.tsx
maricalucian: I had no actual changes in this file, some resolve conflicts went horribly wrong.

PR #127, opened 2019-11-06
update styling

maricalucian:


src/components/account-card/styles.ts
```diff
@@ -1,5 +1,6 @@
 import { StyleSheet, Platform } from 'react-native';
 import { ITheme } from '../../core/theme/itheme';
+import { BASE_DIMENSION } from '../../styles/dimensions';
```
krisboit: this is cool :), we have to make Adrian aware of this, he should use this approach when doing the designs

PR #131, opened 2019-11-07
Qr code reader

georgeburduhos:


src/screens/send/send.tsx
```diff
@@ -55,10 +60,18 @@ export class SendScreenComponent extends React.Component<IProps, IState> {
             amount: '',
             fee: '',
             isValidAddress: false,
-            blockchain: Blockchain.ZILLIQA
+            blockchain: Blockchain.ZILLIQA,
+            openQrCode: false
```
krisboit: this state prop is not used, you can remove it

src/screens/send/send.tsx
```diff
@@ -99,6 +116,9 @@ export class SendScreenComponent extends React.Component<IProps, IState> {
                             this.verifyAddress(text);
                         }}
                     />
+                    <TouchableOpacity onPress={this.onPressQrCodeIcon} style={[styles.qrButton]}>
```
krisboit: you should hide this for web

src/screens/send/styles.ts
```diff
@@ -25,7 +25,7 @@ export default (theme: ITheme) =>
         inputAddress: {
             flex: 1,
             color: theme.colors.text,
-            paddingRight: 40
+            paddingRight: 15
```
krisboit: use the new const @maricalucian added, i think there is no problem if padding would be 16

src/screens/send/styles.ts
```diff
@@ -68,5 +68,13 @@ export default (theme: ITheme) =>
...
             paddingTop: 52,
             backgroundColor: theme.colors.appBackground
+        },
+        icon: {
+            color: theme.colors.accent,
+            marginHorizontal: 0
+        },
+        qrButton: {
+            flexDirection: 'row',
+            marginRight: 5
```
krisboit: same here, make it 4 :)

src/screens/send/styles.ts
```diff
@@ -25,7 +25,7 @@ export default (theme: ITheme) =>
         inputAddress: {
             flex: 1,
             color: theme.colors.text,
-            paddingRight: 40
+            paddingRight: 15
         },
         receipientLabel: {
             paddingLeft: 15,
```
krisboit: same here, make it 16

src/components/qr-modal/qr-modal.tsx
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
+import { View, Modal } from 'react-native';
+import { CameraKitCameraScreen, CameraKitCamera } from 'react-native-camera-kit';
```
krisboit: this is failing for web, since we won't have support for qrcode scan on web just create a new qr-modal.web.tsx and put a dummy empty component. It should solve the problem,

PR #156, opened 2019-11-16
add password modal, delete wallet and unveil mnemonic

maricalucian:


src/components/password-modal/password-modal.tsx
```diff
@@ -0,0 +1,135 @@
...
+        };
+
+        props.onReject && (this.onReject = props.onReject);
+        props.onConfirm && (this.onConfirm = props.onConfirm);
+
+        props.obRef && props.obRef(this);
+    }
+
+    public requestPassword(callback) {
+        this.onConfirm = callback;
```
krisboit: I would make this more simple. Right now if someone uses `onReject` and `onConfirm` props and also calls `requestPassword` `this.onConfirm` is overwritten. I would loose the callback, and i would return a promise:
```ts
    private passwordRequestDeferred;
    public requestPassword(): Promise<string> {
        this.passwordRequestDeferred = new Deferred();
        this.setState({visible: true});
        return this.passwordRequestDeferred.promise;
    }
```
you can find implementation for Deferred here (https://github.com/cryptolandtech/moonlet/blob/master/src/utils/deferred.ts)

on button presses you should add
```ts
// for reject
this.passwordRequestDeferred && this.passwordRequestDeferred.reject();
// for confirm
this.passwordRequestDeferred && this.passwordRequestDeferred.resolve('PASSWORD');
```

krisboit: i would also auto close modal `onReject` and `onConfirm`

src/components/password-modal/password-modal.tsx
```diff
@@ -0,0 +1,135 @@
...
+                            {this.props.buttonLabel}
+                        </Button>
+                    </View>
+                </View>
+            </Modal>
+        );
+    }
+}
+
+export const PasswordModal = withTheme(stylesProvider)(PasswordModalComponent);
```
krisboit: Even if you are using only one enhancer, you could use smartConnect, and add PropTypes, this way we get autocomplete functionality for component props. (See docs)

src/screens/view-wallet-mnemonic/view-wallet-mnemonic.tsx
```diff
@@ -0,0 +1,131 @@
...
+    mnemonic: string[];
+    showPasswordModal: boolean;
+}
+
+export const navigationOptions = ({ navigation }: any) => ({
+    headerLeft: () => {
+        return (
+            <HeaderLeft
+                icon="close"
+                text="Close"
```
krisboit: use translation

src/screens/view-wallet-mnemonic/view-wallet-mnemonic.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+interface IState {
+    mnemonic: string[];
+    showPasswordModal: boolean;
+}
+
+export const navigationOptions = ({ navigation }: any) => ({
+    headerLeft: () => {
+        return (
+            <HeaderLeft
```
krisboit: this should be extracted in a component or a function, as a navigation util.

src/screens/view-wallet-mnemonic/view-wallet-mnemonic.tsx
```diff
@@ -0,0 +1,131 @@
...
+                icon="close"
+                text="Close"
+                onPress={() => {
+                    navigation.goBack(null);
+                }}
+            />
+        );
+    },
+    title: 'View phrase'
```
krisboit: use translation

src/screens/view-wallet-mnemonic/view-wallet-mnemonic.tsx
```diff
@@ -0,0 +1,131 @@
...
+    public async onEnterPassword(password) {
+        const walletId = this.props.navigation.state?.params?.wallet.id;
+
+        if (!walletId) {
+            // TODO show an error?
+            this.props.navigation.goBack(null);
+        }
+
+        try {
+            const passHash = await hash(password);
```
krisboit: i think it would be better to use `HDWallet.loadFromStorage()`, to keep this login in only one place

src/screens/view-wallet-mnemonic/view-wallet-mnemonic.tsx
```diff
@@ -0,0 +1,131 @@
...
+                </View>
+
+                <PasswordModal
+                    buttonLabel={translate('Wallets.unveil')}
+                    infoText={translate('Wallets.unveilPasswordRequest')}
+                    visible={this.state.showPasswordModal}
+                    onReject={() => {
+                        this.props.navigation.goBack(null);
+                    }}
+                    onConfirm={password => {
```
krisboit: for this type of cases, i propose we use `@bind` decorator (https://www.npmjs.com/package/bind-decorator).
the code would look like this:
```ts
@bind
public async onEnterPassword(password) {...}

render() {
  return (...
    onConfirm={this.onEnterPassword}
  ...);
}
```

What do you say ?

src/screens/view-wallet-mnemonic/view-wallet-mnemonic.tsx
```diff
@@ -0,0 +1,131 @@
...
+import stylesProvider from './styles';
+import { withTheme } from '../../core/theme/with-theme';
+import { translate } from '../../core/i18n';
+import { HeaderLeft } from '../../components/header-left/header-left';
+import { smartConnect } from '../../core/utils/smart-connect';
+import { PasswordModal } from '../../components/password-modal/password-modal';
+import { hash } from '../../core/secure/encrypt';
+import { readEncrypted } from '../../core/secure/storage';
+
+export interface IProps {
```
krisboit: use `IThemeProps` and `INavigationProps` instead (see Docs)

src/screens/wallets/wallets.tsx
```diff
@@ -120,26 +136,101 @@ export class WalletsScreenComponent extends React.Component<IProps & IReduxProps
...
+    public onDeleteConfirmed(wallet: IWalletState) {
+        // request password
+        this.setState({
+            showPasswordModal: true
+        });
+
+        this.passwordModalRef.requestPassword(async password => {
+            // verify password by decrypting mnemonic (?)
+            try {
+                const passHash = await hash(password);
```
krisboit: This is ok for HDWallets, but when we introduce Hardware wallets this code will not work, since we won't have the mnemonic for the wallet.  We can leave it as it is for now, but we need to find a solution to check user password.

src/screens/view-wallet-mnemonic/view-wallet-mnemonic.tsx
maricalucian: `HDWallet.loadFromStorage()` does not expose mnemonic and basically uses same function

src/screens/view-wallet-mnemonic/view-wallet-mnemonic.tsx
maricalucian: Cool. Added :)

src/screens/wallets/wallets.tsx
maricalucian: changed to use pass validation in pass modal

src/screens/view-wallet-mnemonic/view-wallet-mnemonic.tsx
maricalucian: added a getter for the above

src/redux/utils/deferred.ts
```diff
@@ -0,0 +1,12 @@
+export class Deferred {
```
krisboit: this should be moved in core folder :)

maricalucian: haha, yup

PR #167, opened 2019-11-19
remove wallet from storage also, add invalid password message

maricalucian:


src/core/secure/storage.ts
```diff
@@ -25,3 +25,11 @@ export const readEncrypted = async (key: string, hash: string) => {
         return Promise.reject(e);
     }
 };
+
+export const deleteWalletFromStorage = async (key: string) => {
```
krisboit: we should rename this, as it's a generic function, i would use `deleteEncrypted` or `deleteFromStorage`

PR #168, opened 2019-11-19
Amount format

georgeburduhos:


src/components/account-card/__tests__/__snapshots__/account-card.test.tsx.snap
```diff
@@ -6,7 +6,7 @@ exports[`account card component renders correctly 1`] = `
   <View style={{...}}>
     <View style={{...}}>
       <withTheme(TextComponent) format={{...}}>
-        12332
+        0.000
```
krisboit: since the test didn't change, and the balance value is mocked to `12332`, this doesn't seem to be fine...

src/core/blockchain/common/account.ts
```diff
@@ -0,0 +1,25 @@
+import { Blockchain } from '../types';
+import { BigNumber } from 'bignumber.js';
+
+export const convertUnit = (
+    blockchain: Blockchain,
+    value: BigNumber,
+    fromUnit,
+    toUnit,
+    info
```
krisboit: this param is not needed, you already have `blockchain`, just import `BLOCKCHAIN_INFO` and `info = BLOCKCHAIN_INFO[blockchain]`

src/core/blockchain/common/account.ts
```diff
@@ -0,0 +1,25 @@
+import { Blockchain } from '../types';
+import { BigNumber } from 'bignumber.js';
+
+export const convertUnit = (
+    blockchain: Blockchain,
+    value: BigNumber,
+    fromUnit,
```
krisboit: try to add types for all parameters where possible

src/core/blockchain/common/account.ts
```diff
@@ -0,0 +1,25 @@
...
+    toUnit,
+    info
+): BigNumber => {
+    if (fromUnit === toUnit) {
+        return value;
+    }
+    const object = new BigNumber(value);
```
krisboit: try to use more suggestive names, in this case `value` or `amount`

src/core/blockchain/types/account.ts
```diff
@@ -9,4 +10,7 @@ export interface IBlockchainAccountUtils {
     privateToPublic(privateKey: string): string;
 
     getAccountFromPrivateKey(privateKey: string, index: number): IAccountState;
+
+    amountToStd(value: string): BigNumber;
```
krisboit: `value` should be of type `BigNumber | number | string`

src/core/i18n/translation/locales/en.ts
```diff
@@ -22,12 +23,16 @@ export const translation: ITranslations = {
                 receive: 'Receive',
                 recover: 'Recover',
                 secure: 'Secure',
+                secureW: 'Secure wallet',
```
krisboit: we shouldn't add screen titles in labels. In labels we should add simple words that are reusable within the application

src/core/i18n/translation/locales/en.ts
```diff
@@ -22,12 +23,16 @@ export const translation: ITranslations = {
...
                 send: 'Send',
+                settings: 'Settings',
                 status: 'Status',
+                tc: 'Terms and conditions',
                 tos: 'Terms of service',
                 transactions: 'Transactions',
                 to: 'To',
-                understand: 'I understand'
+                understand: 'I understand',
+                wallets: 'Wwallets'
```
krisboit: Typo :)

src/core/i18n/translation/locales/en.ts
krisboit: also you could use `secureWallet` as key

src/core/utils/format-amount.ts
```diff
@@ -0,0 +1,24 @@
+import { BigNumber } from 'bignumber.js';
```
krisboit: These are useful functions, but I would simplify things.

Let's create a `Amount` component, that receives these props:
```ts
interface IExternalProps {
    amount: BigNumber | number | string
    currency: string,
    convertTo?: string // if this is present will convert amount to this currency
}
```
So instead of all these functions we can use this component. (Everywhere we need to display an amount, we have the values so there's no need to have specialised function for different kind of objects.

src/screens/dashboard/dashboard.tsx
```diff
@@ -44,23 +44,24 @@ const SCREEN_WIDTH = Dimensions.get('window').width;
...
 
-const calculateBalances = (accounts: IAccountState[]) =>
-    accounts.reduce(
+const calculateBalances = (accounts: IAccountState[]) => {
+    return accounts.reduce(
         (out: any, account: IAccountState) => {
             if (!out.balance[account.blockchain]) {
                 out.balance[account.blockchain] = {
-                    amount: account?.balance?.value || new BigNumber(0)
+                    amount: new BigNumber(account?.balance?.value) || new BigNumber(0)
```
krisboit: instead of doing this in a lot of places we need to fix the core problem. We need to somehow restore correct values in redux.

src/screens/dashboard/dashboard.tsx
```diff
@@ -159,11 +160,16 @@ export class DashboardScreenComponent extends React.Component<IProps & IReduxPro
             '0x1b6c705252438d59DB3ADB85e3B91374377a20c9',
             true
         );
+        this.props.getBalance(
```
krisboit: ok, instead of adding this hardcoded you can get the accounts from redux...

src/screens/transaction-details/transaction-details.tsx
```diff
@@ -54,9 +55,7 @@ export class TransactionDetailsComponent extends React.Component<
         const transaction = this.props.transaction;
         const account = this.props.account;
         const amount =
-            BLOCKCHAIN_INFO[account.blockchain].coin +
-            ' ' +
-            transaction.amount.toString().slice(0, 5);
+            BLOCKCHAIN_INFO[account.blockchain].coin + ' ' + formatAmountFromAccount(account);
```
krisboit: this doesn't seem to be right since amount is the amount of the transaction not account balance.

src/core/blockchain/common/account.ts
georgeburduhos: Problem with this is with circular import. problem discovered when running tests

src/components/amount/styles.ts
```diff
@@ -0,0 +1,5 @@
+import { StyleSheet } from 'react-native';
```
krisboit: i think we can delete this file :)

src/components/coin-balance-card/coin-balance-card.tsx
```diff
@@ -1,14 +1,15 @@
...
-import { Text } from '../../library';
 
 import stylesProvider from './styles';
 import { withTheme } from '../../core/theme/with-theme';
 import BigNumber from 'bignumber.js';
+import { Blockchain } from '../../core/blockchain/types';
+import { Amount } from '../amount/amount';
 
 export interface IProps {
     currency: string;
```
krisboit: as i see, `currency` is not used anymore, you can remove it

src/components/coin-balance-card/coin-balance-card.tsx
```diff
@@ -1,14 +1,15 @@
...
+    blockchain: Blockchain;
     balance: BigNumber;
     toCurrency: string;
```
krisboit: same for `toCurrency`

src/core/blockchain/common/account.ts
```diff
@@ -0,0 +1,24 @@
+import { Blockchain, IBlockchainConfig } from '../types';
+import { BigNumber } from 'bignumber.js';
+
+export const convertUnit = (
+    blockchain: Blockchain,
```
krisboit: `blockchain` is not used, you can delete it

PR #176, opened 2019-11-20
add accounts sreen

maricalucian:


src/redux/wallets/reducer.ts
```diff
@@ -95,6 +102,36 @@ export default (state: IWalletState[] = intialState, action: IAction) => {
...
+        case ACCOUNT_ADD:
+            return state.map(wallet =>
+                wallet.id === action.data.walletId
+                    ? {
+                          ...wallet,
+                          accounts: [].concat(
+                              wallet.accounts,
+                              wallet.accounts.some(
+                                  account =>
+                                      account && account.address === action.data.account.address
```
krisboit: this should be fine, but to be safer you should check the blockchain also

src/redux/wallets/reducer.ts
```diff
@@ -95,6 +102,36 @@ export default (state: IWalletState[] = intialState, action: IAction) => {
...
+                    : wallet
+            );
+
+        case ACCOUNT_REMOVE:
+            return state.map(wallet =>
+                wallet.id === action.data.walletId
+                    ? {
+                          ...wallet,
+                          accounts: wallet.accounts.filter(
+                              account => account.address !== action.data.account.address
```
krisboit: same here, check the blockchain also

src/screens/accounts/accounts.tsx
```diff
@@ -0,0 +1,159 @@
...
+import { Blockchain } from '../../core/blockchain/types';
+import { HDWallet } from '../../core/wallet/hd-wallet/hd-wallet';
+import { selectCurrentWallet } from '../../redux/wallets/selectors';
+import { getPassword } from '../../core/secure/keychain';
+import { formatAddress } from '../../core/utils/format-address';
+import { Text } from '../../library';
+import { Amount } from '../../components/amount/amount';
+import { getBlockchain } from '../../core/blockchain/blockchain-factory';
+
+export interface IProps {
```
krisboit: remove this, as it's not used

src/screens/accounts/accounts.tsx
```diff
@@ -0,0 +1,159 @@
...
+                setAccounts(hdAccounts);
+
+                const client = getBlockchain(props.blockchain).getClient(4);
+
+                const balanceCalls = [];
+                hdAccounts.map(account => {
+                    balanceCalls.push(client.getBalance(account.address));
+                });
+
+                // cache this somehow?
```
krisboit: yeah maybe we should, but i think it's fine for now.

src/screens/dashboard/dashboard.tsx
```diff
@@ -47,6 +48,9 @@ const ANIMATED_BC_SELECTION = true;
 const calculateBalances = (accounts: IAccountState[]) => {
     return accounts.reduce(
         (out: any, account: IAccountState) => {
+            if (!account) {
```
krisboit: why do we need this check ?

maricalucian: if somehow (from some error?) a null value gets in accounts state, app wont load anymore. basically this should never happen, we could remove it

krisboit: ok, i understand :)

PR #177, opened 2019-11-21
Keyboard

NovacAndrei:


src/components/keyboard-custom/words.js
```diff
@@ -0,0 +1,47 @@
...
+    { lower: '4' },
+    { lower: '5' },
+    { lower: '6' },
+    { lower: '7' },
+    { lower: '8' },
+    { lower: '9' },
+    { lower: '0' }
+];
+
+export const firstRow = [
```
krisboit: there's no need to keep lower and upper cases, you can use `.toLowercase()` and `.toUppercase()` :)
so all these consts can be one array :)
ex:
```ts
const keyboardLayout = [
[1,2,3,4,5,6,7,8,9,0],
['q', 'w', 'e', ...],
 ...
];
```

Also out project is using typescript, all files in `/src` should be with `.ts` or `.tsx` extensions

krisboit: with these changes you can keep the const in component file :)

PR #178, opened 2019-11-21
Splash Screen

NovacAndrei:
- lottie


src/app.tsx
```diff
@@ -9,24 +9,30 @@ import { darkTheme } from './styles/themes/dark-theme';
...
     //    public unsubscribe =
     private translationsLoaded: boolean = false;
     private reduxStateLoaded: boolean = false;
 
     constructor(props: any) {
         super(props);
         this.state = {
-            appReady: false
+            appReady: false,
+            defaultTimer: 1
```
krisboit: Let's keep it simple.
You should use `setTimeout()` not `setInterval()`. In state keep a property `splashAnimationDone: boolean`.

Example: 

```ts
constructor(props: any) {
  this.state = {
    appReady: false,
    splashAnimationDone: false
  }
  setTimeout(() => this.setState({splashAnimationDone: true}); // you can move this in componentDidMount()
}

render() {
  if (this.state.appReady && this.state.splashAnimationDone) {
    // all done, render app
  } else {
    // render splash
  }
}
```

krisboit:
we need to do one more thing, we have to clear ios launchscreen, and set a background color, as our app

PR #183, opened 2019-11-22
keyboard onTouchStart

NovacAndrei:


src/components/keyboard-custom/keyboard-custom.tsx
```diff
@@ -34,6 +34,10 @@ export class KeyboardComponent extends React.Component<
         };
     }
 
+    public onTouchStart = (currentWord: any) => {
```
krisboit: please don't use arrow function sin classes. 

Also you can use `@bind` decorator so `this.onTouchStart.bind(this)` is not necessary

PR #194, opened 2019-11-26
Remove app tokens

georgeburduhos:


krisboit:
Did you regenerate tokens before adding them to Github Secrets ? Also make sure old token are invalidated.

PR #197, opened 2019-11-26
Custom Keyboard

NovacAndrei:


.github/workflows/main.yml
```diff
@@ -2,6 +2,7 @@ on:
   push:
     branches:
       - master
+      - "keyboard-fix"
```
krisboit: this should be removed before merging in master

PR #200, opened 2019-11-28
Address Book

NovacAndrei:


src/screens/send/components/address-book/address-book.tsx
```diff
@@ -0,0 +1,211 @@
...
+const mapDispatchToProps = {
+    deleteContact,
+    updateContactName
+};
+
+export class AddressBookComponent extends React.Component<
+    IReduxProps & IExternalProps & IThemeProps<ReturnType<typeof stylesProvider>>,
+    IState
+> {
+    public walletSwipableRef: any[] = new Array();
```
krisboit: The name of the attribute is confusing. Use a more suggestive name, like `contactsSwipeableRefs`

src/screens/send/components/address-book/address-book.tsx
```diff
@@ -0,0 +1,211 @@
...
+        return (
+            <Swipeable
+                key={index}
+                ref={ref => {
+                    this.walletSwipableRef[index] = ref;
+                }}
+                renderLeftActions={() => this.renderLeftActions(contact)}
+                onSwipeableWillOpen={() => {
+                    if (
+                        index !== this.state.openedSwipeIndex &&
```
krisboit: since `openedSwipeIndex` doesn't affect rendering directly you can move it from state to class attribute. Every time you do a setState and something changes, you trigger a new render cycle and, in this case, you don't need that.

src/screens/send/components/address-book/address-book.tsx
```diff
@@ -0,0 +1,211 @@
...
+                        this.walletSwipableRef[this.state.openedSwipeIndex]
+                    ) {
+                        this.walletSwipableRef[this.state.openedSwipeIndex].close();
```
krisboit: you already have `closeCurrentOpenedSwipable()`, use it!

src/screens/send/components/address-book/address-book.tsx
```diff
@@ -0,0 +1,211 @@
...
+                    }
+                    this.setState({ openedSwipeIndex: index });
+                }}
+            >
+                <TouchableOpacity
+                    style={styles.rowContainer}
+                    onPress={() => {
+                        const foundAccount = this.props.accounts.find(
```
krisboit: Why are you doing this ? Contact addresses are usually different from user accounts addresses, It the address of a friend...

src/screens/send/components/address-book/address-book.tsx
```diff
@@ -0,0 +1,211 @@
...
+import { formatAddress } from '../../../../core/utils/format-address';
+import { translate } from '../../../../core/i18n';
+import { Blockchain } from '../../../../core/blockchain/types';
+import Swipeable from 'react-native-gesture-handler/Swipeable';
+import { deleteContact, updateContactName } from '../../../../redux/contacts/actions';
+import { confirm } from '../../../../core/utils/dialog';
+import { IAccountState } from '../../../../redux/wallets/state';
+
+export interface IReduxProps {
+    contacts: IContactsState[];
```
krisboit: the return type of `selectContacts` selector is not a `IContactsState[]`, you have 2 options:
1. The easy one, `any[]`
2. The nice one, `ReadonlyArray<SectionListData< IContactsState >>`

src/screens/send/components/address-book/address-book.tsx
```diff
@@ -0,0 +1,211 @@
...
+                    <Text style={styles.addAddressBookText}>
+                        {translate('Send.addAddressBook')}
+                    </Text>
+                </View>
+            );
+        } else {
+            return (
+                <SafeAreaView style={styles.container}>
+                    <SectionList
+                        sections={contacts as []}
```
krisboit: if you use the correct type for contacts prop, you don't need this casting here

src/screens/send/components/address-book/address-book.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+export interface IReduxProps {
+    contacts: IContactsState[];
+    deleteContact: typeof deleteContact;
+    updateContactName: typeof updateContactName;
+}
+
+export interface IExternalProps {
+    blockchain: Blockchain;
+    accounts: IAccountState[];
```
krisboit: you don't really need user accounts here, you only have to display the contacts, and when a contacts is selected return the address an maybe the name of contact

src/screens/send/components/address-book/address-book.tsx
```diff
@@ -0,0 +1,211 @@
...
+    onAccountSelection: (account: IAccountState) => any;
```
krisboit: this should be `onContactSelected(contact: IContactState)`

src/core/utils/dialog.ts
```diff
@@ -0,0 +1,19 @@
+import { Alert } from 'react-native';
+
+export const confirm = (title: string, message: string): Promise<void> => {
```
krisboit: seeing this in action, made me think how we can make it easier to use. Right now we would always have to handle promise rejection, wich is ugly :)

A better way would be to return a `Promise<boolean>` and always resolve the promise:
- with `True` if used clicked OK
- with `False` if user clicked Cancel

This way we could use this function like this:
```ts
if (await confirm('Title', 'Message')) {...}
```
we can do an exception and not use try..catch in this case, we know that `confirm` will always resolve values.

src/screens/send/send.tsx
```diff
@@ -35,23 +30,31 @@ import bind from 'bind-decorator';
...
+    wallet: IWalletState;
+    isContactAlreadySaved: typeof isContactAlreadySaved;
 }
 
 export const mapStateToProps = (state: IReduxState, ownProps: INavigationParams) => {
     return {
         account: getAccount(state, ownProps.accountIndex, ownProps.blockchain),
-        accounts: getAccounts(state, ownProps.blockchain)
+        accounts: getAccounts(state, ownProps.blockchain),
+        wallet: selectCurrentWallet(state)
```
krisboit: this is not used, delete it.

src/screens/send/components/address-book/address-book.tsx
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    public renderLeftActions = (contact: IContactState) => {
+        const styles = this.props.styles;
+        return (
+            <View style={styles.leftActionsContainer}>
+                <TouchableOpacity
+                    style={styles.action}
+                    onPress={() => {
+                        confirm(translate('Send.deleteContact'), '')
```
krisboit: you can move the confirm in `deleteContact` action creator, so from here you only call `this.props.deleteContact(contact);`. This way this would be more reusable.

src/screens/send/send.tsx
```diff
@@ -35,23 +30,31 @@ import bind from 'bind-decorator';
 
 export interface IProps {
     navigation: NavigationScreenProp<NavigationState, NavigationParams>;
```
krisboit: this is not needed anymore, `INavigationProps` is used

src/redux/contacts/actions.ts
```diff
@@ -0,0 +1,37 @@
...
+        data: contactData
+    };
+};
+
+export const isContactAlreadySaved = (address: string) => (
+    dispatch: Dispatch,
+    getState: () => IReduxState
+) => {
+    const state = getState();
+    return Object.keys(state.contacts).some(c => c === address);
```
krisboit: Also to achieve this you should do this:
```ts
return !!state.contact[`${blockchain}|${address}`];
```

src/screens/send/send.tsx
```diff
@@ -273,81 +331,92 @@ export class SendScreenComponent extends React.Component<
...
-                            />
-                        )}
-
-                        <QrModalReader
-                            ref={ref => (this.qrCodeScanner = ref)}
-                            onQrCodeScanned={this.onQrCodeScanned}
+                    </TouchableOpacity>
+
+                    {this.state.isValidAddress &&
+                        !this.props.isContactAlreadySaved(this.state.toAddress) &&
```
krisboit: you don't really need a function here, you just need the contacts and you can do this check
```tsx
{this.state.isValidAddress &&
!this.props.contacts[`${this.props.account.blockchain}|${this.state.toAddress}`] && ... }
```

Also i would move the check in `renderAddAddressToBook`

PR #202, opened 2019-11-29
Edit wallet name

NovacAndrei:


src/redux/wallets/actions.ts
```diff
@@ -231,3 +232,10 @@ export const deleteWallet = walletId => (
 
     deleteFromStorage(walletId);
 };
+
+export const updateWAlletName = (walletId: string, newName: string) => {
```
krisboit: typo use `updateWalletName`

PR #206, opened 2019-12-02
Header height, icon size and target size

NovacAndrei:


src/styles/dimensions.ts
```diff
@@ -1,2 +1,4 @@
 export const BORDER_RADIUS = 6;
 export const BASE_DIMENSION = 8;
+export const ICON_SIZE = 24;
+export const ICON_CONTAINER = 44;
```
krisboit: the name is too generic, `ICON_CONTAINER_SIZE` would be more suitable

PR #207, opened 2019-12-03
Balances carousel

NovacAndrei:


NovacAndrei:
Balance is taken from blockchain

```
const balance = await getBlockchain(blockchain)
                .getClient(chainId)
                .getBalance(address);
```

PR #208, opened 2019-12-03
Settings - network selection

NovacAndrei:


src/core/i18n/translation/locales/en.ts
```diff
@@ -144,6 +144,31 @@ export const translation: ITranslations = {
         },
         Watch: {
             title: 'Watch'
+        },
+        Settings: {
+            title: 'Settings',
+            security: 'SECURITY',
```
krisboit: texts like this, one word, can be moved to `App.labels`, and you can use `Security`, where you use it you can use `textTransform: 'uppercase'` style on element

PR #217, opened 2019-12-06
Settings - Blockchain portfolio

NovacAndrei:


src/redux/app/actions.ts
```diff
@@ -25,9 +29,23 @@ export const toggleTestNet = () => {
...
 
-export const setNetworkTestNetChainId = (blockchain: string, chainId: number) => {
+export const setNetworkTestNetChainId = (blockchain: Blockchain, chainId: number) => {
     return {
         type: APP_SET_NETWORK_TEST_NET_CHAIN_ID,
         data: { blockchain, chainId }
     };
 };
+
+export const toggleNetwork = (blockchain: Blockchain) => {
```
krisboit: `toogleBlockchainActive`

src/redux/app/actions.ts
```diff
@@ -25,9 +29,23 @@ export const toggleTestNet = () => {
...
+    return {
+        type: APP_TOGGLE_NETWORK,
+        data: { blockchain }
+    };
+};
+
+export const sortNetworks = (sortedNetworks: Array<{ blockchain: Blockchain; order: number }>) => {
```
krisboit: `updateBlockchianOrder`

src/redux/app/actions.ts
```diff
@@ -1,8 +1,12 @@
+import { Blockchain } from '../../core/blockchain/types';
+
 // actions consts
 export const APP_SWITCH_WALLET = 'APP_STATE_SWITCH_WALLET';
 export const APP_SET_TOS_VERSION = 'APP_SET_TOS_VERSION';
 export const APP_SET_TEST_NET = 'APP_SET_TEST_NET';
 export const APP_SET_NETWORK_TEST_NET_CHAIN_ID = 'APP_SET_NETWORK_TEST_NET_CHAIN_ID';
+export const APP_TOGGLE_NETWORK = 'APP_TOGGLE_NETWORK';
```
krisboit: `APP_TOGGLE_BLOCKCHAIN`

src/redux/app/actions.ts
```diff
@@ -1,8 +1,12 @@
...
+export const APP_SORT_NETWORKS = 'APP_SORT_NETWORKS';
```
krisboit: `APP_UPDATE_BLOCKCHAIN_ORDER`

src/redux/app/actions.ts
krisboit: i think you can use `Array<Blockchain>`, as it is now i expect values to be:
```ts
[
    {blockchain: 'ZILLIQA', order: 0},
    {blockchain: 'ETHEREUM', order: 1}
]
```
`order` is the same with element index

src/redux/app/reducer.ts
```diff
@@ -15,10 +17,14 @@ const intialState: IAppState = {
     testNet: true,
     networks: {
         [Blockchain.ETHEREUM]: {
```
krisboit: we have to decide if the ordering affects all wallets or we have ordering for each wallet separatly

krisboit: also we should add a new property `blockchains` on the same level with `networks`. `networks` should have only networks info.

src/screens/settings/blockchain-portfolio/blockchain-portfolio.tsx
```diff
@@ -0,0 +1,108 @@
...
+        backgroundColor: themes[theme].colors.header
+    }
+});
+
+export class BlockchainPortfolioComponent extends React.Component<
+    INavigationProps & IReduxProps & IThemeProps<ReturnType<typeof stylesProvider>>
+> {
+    public static navigationOptions = navigationOptions;
+
+    public renderNetwork = ({ item, index, move, moveEnd, isActive }: any) => {
```
krisboit: please don't use arrow functions in classes
we definitely need tslint rule :)

src/screens/settings/blockchain-portfolio/blockchain-portfolio.tsx
```diff
@@ -0,0 +1,108 @@
...
+        return (
+            <View style={styles.container}>
+                <DraggableFlatList
+                    data={this.props.networks}
+                    renderItem={this.renderNetwork}
+                    keyExtractor={(item: any) => `${item.value.order}`}
+                    scrollPercent={5}
+                    onMoveEnd={({ data }) =>
+                        this.props.sortNetworks(
+                            data.reduce((array, blockchain, index) => {
```
krisboit: i think you can use `map()` here

src/screens/settings/blockchain-portfolio/blockchain-portfolio.tsx
```diff
@@ -0,0 +1,108 @@
...
+}
+
+const mapDispatchToProps = {
+    toggleNetwork,
+    sortNetworks
+};
+
+const mapStateToProps = (state: IReduxState) => {
+    const networks = Object.keys(state.app.networks)
+        .reduce((array, key) => {
```
krisboit: use `map()` instead

src/screens/settings/blockchain-portfolio/blockchain-portfolio.tsx
```diff
@@ -0,0 +1,108 @@
...
+    sortNetworks: typeof sortNetworks;
+}
+
+const mapDispatchToProps = {
+    toggleNetwork,
+    sortNetworks
+};
+
+const mapStateToProps = (state: IReduxState) => {
+    const networks = Object.keys(state.app.networks)
```
krisboit: `networks` type is different from the type of `networks` prop.

src/components/splash-screen/styles.ts
```diff
@@ -15,6 +15,6 @@ export default (theme: ITheme) =>
         lottie: {
             ...StyleSheet.absoluteFillObject,
             height: '100%',
-            top: -150
+            top: -BASE_DIMENSION * 10
```
maricalucian: This dimension needs to be exact? Is needed so the animation exists screen to the top?
If not, is better to specify larger dimension in percents.

src/screens/settings/blockchain-portfolio/blockchain-portfolio.tsx
```diff
@@ -9,31 +9,27 @@ import { Icon } from '../../../components/icon';
...
-    const networks = Object.keys(state.app.networks)
-        .reduce((array, key) => {
-            return [...array, { key, value: state.app.networks[key] }];
-        }, [])
-        .sort((a, b) => a.value.order - b.value.order);
-
     return {
-        networks
+        blockchains: Object.keys(state.app.blockchains)
+            .reduce((array, key) => [...array, { key, value: state.app.blockchains[key] }], [])
```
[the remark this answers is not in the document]
NovacAndrei: Changed to `map`, easier to understand this way

.github/workflows/main.yml
```diff
@@ -46,8 +46,6 @@ jobs:
       - uses: actions/setup-ruby@v1
         with:
           ruby-version: "2.x"
-      - name: Generate firebase config file
```
krisboit: this shouldn't be removed

.github/workflows/main.yml
```diff
@@ -88,8 +86,8 @@ jobs:
           ssh-private-key: ${{ secrets.SSH_CERTS }}
       - name: Add to host
         run: ssh-keyscan gitlab.com >> ~/.ssh/known_hosts
-      - name: Generate firebase config file
```
krisboit: same here

.github/workflows/main.yml
```diff
@@ -88,8 +86,8 @@ jobs:
...
-        run: echo '${{ secrets.FB_CONFIG_IOS_BETA }}' > ios/GoogleService-Info.plist 
+      - name: Checkout
```
krisboit: this shouldn't be added back

PR #246, opened 2019-12-11
Account screen request pin on reveal private key

NovacAndrei:


src/screens/account/components/account-settings/account-settings.tsx
```diff
@@ -27,34 +28,42 @@ interface IState {
...
-            showBackButton: true,
-            title: translate('AccountSettings.revealPrivate'),
-            key: this.props.account.address // TO DO - switch to private key
-        });
+        this.passwordModal.requestPassword().then(() =>
+            this.setState({
+                showKeyScreen: true,
+                showBackButton: true,
+                title: translate('AccountSettings.revealPrivate'),
+                key: this.props.account.address, // TO DO - switch to private key
```
krisboit: it seems that the privateKey is wrong, I will create a new task to fix this.

PR #248, opened 2019-12-12
Ask pin when app is minimised

NovacAndrei:


src/app.tsx
```diff
@@ -98,10 +99,10 @@ export default class App extends React.Component<{}, IState> {
         AppState.removeEventListener('change', this.handleAppStateChange);
     }
 
-    public requestPassword() {
-        this.passwordModal
-            .requestPassword()
-            .then(() => this.setState({ appState: APP_STATE_ACTIVE }));
+    public showPasswordModal() {
+        this.setState({ showPasswordModal: true }, () =>
+            this.setState({ showPasswordModal: false, appState: APP_STATE_ACTIVE })
```
krisboit: why do we need this ?

NovacAndrei: Otherwise `PaswordModal` won't be visible when app comes back from background

```
public componentDidUpdate(prevProps: IExternalProps) {
        if (this.props.visible !== prevProps.visible && this.props.visible) {
            this.setState({ visible: this.props.visible });
        }
    }
```

Do you have any suggestions how to adjust this?

PR #250, opened 2019-12-12
Disable taking screenshots

NovacAndrei:
- Working only for Android


src/core/utils/prevent-screenshot.ts
```diff
@@ -0,0 +1,17 @@
+import { NativeModules } from 'react-native';
+
+export const forbidFunction = async () => {
```
krisboit: you should rename this to `forbidScreenshots`

src/core/utils/prevent-screenshot.ts
```diff
@@ -0,0 +1,17 @@
...
+    try {
+        await NativeModules.PreventScreenshotModule.forbid();
+    } catch (e) {
+        //
+    }
+};
+
+export const allowFunction = async () => {
```
krisboit: you should rename this to `allowScreenshots`

src/core/utils/prevent-screenshot.ts
```diff
@@ -0,0 +1,17 @@
+import { NativeModules } from 'react-native';
```
krisboit: also the file name should be `screenshot` since it also has an allow function :P

src/screens/account/components/view-key/view-key.tsx
```diff
@@ -32,6 +33,12 @@ export class ViewKeyComponent extends React.Component<
         this.state = {
             copied: false
         };
+
+        forbidFunction();
```
krisboit: well, this is weird :) forbidFunction could mean anything in this context :)
but using `forbidScreenshot` it's more self explanatory

krisboit:
it would be interesting to transform this in a library :)

NovacAndrei:
sure, I'll transform this in a library :)

PR #264, opened 2019-12-15
Add wallet connect

maricalucian:


package.json
```diff
@@ -26,12 +26,15 @@
...
         "@types/react-native-vector-icons": "^6.4.4",
+        "@walletconnect/browser": "^1.0.0-beta.39",
+        "@walletconnect/react-native": "^1.0.0-beta.39",
         "@zilliqa-js/account": "^0.8.4",
         "@zilliqa-js/crypto": "^0.8.1-alpha.0",
         "@zilliqa-js/util": "^0.8.3",
         "bignumber.js": "^9.0.0",
         "bind-decorator": "^1.0.11",
         "bip39": "^3.0.2",
+        "can-promise": "0.0.1",
```
krisboit: it seems this is not used. Is it needed ?

src/screens/watch/watch.tsx
```diff
@@ -29,6 +35,14 @@ export class WatchScreenComponent extends React.Component<
         Clipboard.setString(getApnsToken());
     };
 
+    public onQrCodeScanned = async (value: string) => {
+        WalletConnectClient.connect(value);
+        const keychainPassword = await getPassword();
+        if (keychainPassword) {
+            storeEncrypted(value, WC_CONNECTION_STRING, keychainPassword.password);
```
krisboit: i think this is not quite ok, we need to store the session not the connection string. I think that if we try to reconnect with connection string and the other party is not connected the handshake cannot be done. If we store the session we just connect back, no handshake is required.

PR #268, opened 2019-12-19
Redux Wallet from array to object

NovacAndrei:


src/redux/wallets/actions.ts
```diff
@@ -79,22 +79,20 @@ export const createHWWallet = (
...
+            id: walletId,
+            hwOptions: {
+                deviceId,
+                deviceVendor,
+                deviceModel,
+                connectionType
+            },
+            name: `Wallet ${Object.keys(getState().wallets).length + 1}`,
+            type: WalletType.HW,
+            accounts: account.reduce((out, accounts) => out.concat(accounts), [])
```
krisboit: it seems that the reduce concatenates accounts, in this case i think we can use accounts: account` or am I missing something ?

NovacAndrei: image

It seems that `accounts` it's an array of `IAccountState` arrays and the reduce pulls out into an array of objects.

krisboit: that's true, but i think you tried `createHDWallet` not `createHWWallet`.
In `createHWWallet` we have `const account = await wallet.getAccounts(blockchain, 0);`, in this case `account` should be an array of `IAccountState`.

NovacAndrei: Yes indeed, I double checked and this is correct.`account` is basically `IAccountState[]`. I was looking at `createHDWallet`. I have updated to code. Sorry for the mistake.

PR #273, opened 2020-01-07
Fix Design - v1.1

NovacAndrei:
- change colors
- bottom sheet style colors
- FaceID: ios - password pin and settings screen FaceID option
- copy mnemonic (create wallet mnemonic screen)


src/screens/send/components/gas-fee-advanced/styles.ts
```diff
@@ -15,7 +15,7 @@ export default (theme: ITheme) =>
             borderRadius: BORDER_RADIUS,
             borderColor: 'gray',
             alignSelf: 'stretch',
-            backgroundColor: theme.colors.inputBackground,
```
krisboit: i think it's better to have different vars for input bg color and text colors. For dark theme  maybe they have the same value but on other themes it's not the case.

src/styles/colors.ts
```diff
@@ -1,17 +1,28 @@
...
+    BLACK: '#000000',
+
+    AQUA: '#00DAFF', // main color
+    LIGHT_AQUA: '#00515E', // cards outline
+
+    WHITE_87: '#E0E0E0', // headers from page
+    WHITE_67: '#B1B1B1', // texts from page
+    WHITE_37: '#6A6A6A', // texts from fields
+
+    GRAY_CARDS: '#28282A', // all cards
```
krisboit: color names should not be based on features. you can use https://www.color-blindness.com/color-name-hue/ to get names for different kind of colors

PR #274, opened 2020-01-08
Ledger sign

georgeburduhos:


src/core/wallet/hw-wallet/ledger/apps/zil-interface.js
```diff
@@ -109,7 +109,7 @@ class Zilliqa {
             null,
             2
         );
-        // console.log(chalk.green(message));
+        console.log('tx params4 ', message);
```
krisboit: this should be a .ts file :)

PR #302, opened 2020-01-22
Redux updates

georgeburduhos:


src/redux/wallets/actions.ts
```diff
@@ -51,19 +54,43 @@ export const addWallet = (walletData: IWalletState) => {
...
+    dispatch({
+        type: WALLET_SELECT_BLOCKCHAIN,
+        data: {
+            walletId: wallet.id,
+            blockchain
+        }
+    });
+};
+
+export const switchSelectedAccount = (selectedAccount: ISelectedAccount) => (
```
krisboit: let's use the same naming pattern, `setSelectedAccount`

PR #306, opened 2020-01-29
remote config

georgeburduhos:


src/core/utils/remote-feature-config.ts
```diff
@@ -0,0 +1,41 @@
...
+import DeviceInfo from 'react-native-device-info';
+
+export enum REMOTE_FEATURE {
+    NEAR = 'feature_near'
+}
+
+export const getRemoteConfigFeatures = async () => {
+    // TODO: decide if 1 hour should be the optimal fetch duration
+    // set fetch cache duration to 1 hour for live environment
+    let duration = 3600;
```
krisboit: since this is not for production, let's use a lower value, like 15 mins

src/core/utils/remote-feature-config.ts
```diff
@@ -0,0 +1,41 @@
+import firebase from 'react-native-firebase';
+import DeviceInfo from 'react-native-device-info';
+
+export enum REMOTE_FEATURE {
```
krisboit: enum should be in camel case format, `RemoteFeature`

src/core/utils/remote-feature-config.ts
```diff
@@ -0,0 +1,41 @@
...
+    const data = {};
+    // Retrieve values
+    Object.keys(objects).forEach(key => {
+        data[key] = objects[key].val();
+    });
+
+    return data;
+};
+
+export const isFeatureActive = async (feature: REMOTE_FEATURE): Promise<boolean> => {
```
krisboit: this shoudn't be async, we could call `getRemoteConfigFeatures` on app bootstrap and cache the data, after that `isFeatureActive` could use cached data. It's easier and more user friendly this way.

krisboit: by doing this function sync, it will be easier to use it in components.

PR #309, opened 2020-02-03
Refactor dashboard

georgeburduhos:


src/redux/wallets/reducer.ts
```diff
@@ -106,10 +106,11 @@ export default (state: IWalletsState = intialState, action: IAction) => {
                 [action.data.walletId]: {
                     ...state[action.data.walletId],
                     accounts: state[action.data.walletId].accounts.map(account => {
-                        account.index === action.data.selectedAccount.index &&
                         account.blockchain === action.data.selectedAccount.blockchain
```
krisboit: id would be more clear to have it like this:
```ts
if (account.blockchain === action.data.selectedAccount.blockchain) {
    account.selected = account.index === action.data.selectedAccount.index;
}
```
Let's not use ternary operator as an if replacement. For variable declaration it's fine.

PR #311, opened 2020-02-04
NEAR Bugs

NovacAndrei:


src/redux/wallets/selectors.ts
```diff
@@ -78,6 +78,20 @@ export const getAccounts = (state: IReduxState, blockchain: Blockchain): IAccoun
     return accounts;
 };
 
+export const selectedBlockchainHasAccounts = (state: IReduxState): boolean => {
```
krisboit: i think it would be more useful to have a selector: `getSelectedBlockchainAccounts(state): IAccountState[]` that will return the accounts of the selected blockchain. Your check will become:
`getSelectedBlockchainAccounts().length > 0`

src/core/blockchain/types/config.ts
```diff
@@ -35,5 +35,8 @@ export interface IBlockchainConfig {
     };
     ui: {
         addressDisplay?: 'stripped';
+        enableTokenManagement: boolean;
+        enableAccountCreation?: boolean;
+        maxAccountsNumber?: number;
```
krisboit: this shouldn't be optional :P

PR #334, opened 2020-02-05
Near: send screen all balance button not working properly

NovacAndrei:
Fix gas price and gas limit


src/screens/send/components/fee-total/fee-total.tsx
```diff
@@ -30,6 +30,7 @@ export class FeeTotalComponent extends React.Component<
                         blockchain={this.props.blockchain}
                         token={this.props.token.symbol}
                         tokenDecimals={this.props.token.decimals}
+                        uiDecimals={20} // TODO: remove this
```
krisboit: you should remove this :)

PR #344, opened 2020-02-06
Bug fixing

NovacAndrei:
- Accounts bottom sheet close on selection #320
- PIN Login hide fingerprint icon when touch id is not active #324
- Manage account screen: user shouldn't be able to deactivate a native token #329

### DIFF
--- a/src/components/password-modal/components/password-pin/password-pin.tsx
+++ b/src/components/password-modal/components/password-pin/password-pin.tsx
@@ -267,23 +267,27 @@ export class PasswordPinComponent extends React.Component<
 
     public renderFooterRow = () => {
         const styles = this.props.styles;
+        const icon =
+            Platform.OS === 'ios' && this.state.biometryType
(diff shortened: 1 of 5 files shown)


src/components/password-modal/components/password-pin/password-pin.tsx
```diff
@@ -267,23 +267,27 @@ export class PasswordPinComponent extends React.Component<
 
     public renderFooterRow = () => {
         const styles = this.props.styles;
+        const icon =
+            Platform.OS === 'ios' && this.state.biometryType
```
krisboit: this condition is way too complicated :))
i think it would be more simple to do this:
```
const icon = Platform.OS === 'ios' && this.state.biometryType === 'FaceID' ? 'face-id' : 'touch-id';
```

NovacAndrei: Yes indeed :)) Fixed

PR #360, opened 2020-02-10
Name service

georgeburduhos:


src/core/blockchain/near/name-service.ts
```diff
@@ -0,0 +1,35 @@
+import {
+    IResolveTextResponse,
+    IResolveNameResponse,
+    ResolveTextType,
+    ResolveTextCode
+} from '../types';
+import { Client as NearClient, Client } from './client';
+
+export class NameService {
```
krisboit: this class should implement `IBlockchainNameServiceUtils ` interface

src/core/blockchain/types/name-service.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export interface IResolveTextResponse {
+    code: ResolveTextCode;
+    type: ResolveTextType;
+    address: string;
+    name: string;
+}
+
+export interface IBlockchainNameServiceUtils {
```
krisboit: we can rename this to `IBlockchainNameService`

src/core/blockchain/zilliqa/name-service.ts
```diff
@@ -0,0 +1,33 @@
+import {
+    IResolveTextResponse,
+    IResolveNameResponse,
+    ResolveTextType,
+    ResolveTextCode,
+    ResolveTextError
+} from '../types';
+import { isValidAddress, isValidChecksumAddress } from './account';
+export class NameService {
```
krisboit: this class should implement `IBlockchainNameServiceUtils ` interface

src/screens/send/send.tsx
```diff
@@ -180,35 +192,80 @@ export class SendScreenComponent extends React.Component<
...
+                }
+                case ResolveTextCode.WARN_CHECKSUM: {
+                    this.setState({
+                        isValidText: true,
+                        errorResponseText: undefined,
+                        warningResponseText: translate('Send.receipientWarning')
+                    });
+                    break;
+                }
+                default:
```
krisboit: we should handle this case somehow, maybe as an error

src/screens/send/send.tsx
```diff
@@ -180,35 +192,80 @@ export class SendScreenComponent extends React.Component<
...
+                    break;
+            }
+        } catch (error) {
+            switch (error.error) {
+                case ResolveTextError.INVALID: {
```
krisboit: we should handle connection errors here also

PR #376, opened 2020-02-19
Send page, validate input fields as numeric

NovacAndrei:


src/core/i18n/translation/locales/en.ts
```diff
@@ -123,7 +124,9 @@ export const translation: ITranslations = {
         Fee: {
             feeTitle: 'Transaction fee',
             gasPrice: 'Gas price',
-            gasLimit: 'Gas limit'
+            gasLimit: 'Gas limit',
+            errorGasPrice: 'Invalid Gas Price',
+            errorLimitPrice: 'Invalid Limit Price'
```
georgeburduhos: Invalid Gas Limit Price should be here

src/screens/send/send.tsx
```diff
@@ -368,6 +368,7 @@ export class SendScreenComponent extends React.Component<
     }
 
     public async alertModalAddAddress() {
+        // TODO: check this, it's not opening all the time
```
georgeburduhos: same problem as on Delete wallet?

NovacAndrei: Same problem, yes. The modal does not open all the time.

NovacAndrei: I had also tried with `await delay(0)`, but it didn't help.

src/core/utils/format-number.ts
```diff
@@ -47,3 +47,6 @@ export const formatNumber = (amount: number | BigNumber, options: INumberFormatO
         ? formattedNumber + ' ' + options.currency
         : formattedNumber;
 };
+
+const numericRegex = new RegExp(/^[+-]?\d+(\.\d+)?$/);
```
krisboit: is this what we want ? you should understand regex expressions you use :P
try https://regexr.com/

krisboit: try to understand how this regex rule works, after that i will tell you how to do it without regex :)

PR #377, opened 2020-02-19
ZRC2 Integration

NovacAndrei:


src/core/blockchain/zilliqa/client.ts
```diff
@@ -103,6 +107,26 @@ export class Client extends BlockchainGenericClient {
...
+        return this.call('GetSmartContractSubState', [
+            fromBech32Address(contractAddress)
+                .replace('0x', '')
+                .toLowerCase(),
+            field,
+            subFields
+        ]).then(response => response?.result);
+    }
+
+    public async getSmartContractInit(address: string) {
```
krisboit: this method could be used with both formats, i would do this:
- check if address is a bech32 address
  - if yes call fromBech32
- remove 0x and call toLowerCase()

src/core/blockchain/zilliqa/tokens/zrc2-client.ts
```diff
@@ -0,0 +1,65 @@
...
+        return this.findSmartContractSubField(smartContractInit, 'name');
+    }
+
+    public async getDecimals(contractAddress) {
+        const smartContractInit = await this.getTokenDetails(contractAddress);
+        return this.findSmartContractSubField(smartContractInit, 'decimals');
+    }
+
+    public async getTokenInfo(contractAddress) {
+        const info = await Promise.all([
```
krisboit: with this implementation you are doing 6 requests, but only 2 are needed. Merge this method with `getTokenDetails()` and in `getDecimals`, `getName `, `getSymbol` use that implementation. Example:
```ts
getTokenDetails(...) {
...
}

getDecimals(contractAddress) {
    let info = await this.getTokenDetails(contractAddress);
    return info.decimals;
}
```

src/core/blockchain/zilliqa/transaction.ts
```diff
@@ -40,9 +41,29 @@ export const sign = async (tx: IBlockchainTransaction, privateKey: string): Prom
         gasLimit: Long.fromString(tx.feeOptions.gasLimit.toString()),
         code: '',
         data: '',
-        signature: ''
+        signature: '',
+        priority: true
     };
 
+    // TODO: this should be moved in buildTransferTransaction on data.raw field
```
krisboit: this should be moved in buildTransferTransaction on data.raw field

PR #383, opened 2020-02-24
Fix all balance

NovacAndrei:


src/screens/send/send.tsx
```diff
@@ -291,9 +291,9 @@ export class SendScreenComponent extends React.Component<
     };
 
     public onAddAllBalance = () => {
-        const token = this.props.account.tokens[
-            getBlockchain(this.props.account.blockchain).config.coin
-        ];
+        const { symbol } = this.props.navigation.state.params.token;
```
krisboit: navigation state is mapped into props. use `this.props.token` instead.

src/screens/send/__tests__/__snapshots__/send.test.tsx.snap
```diff
@@ -5,7 +5,7 @@ exports[`SendScreen Buton should open qr-code 1`] = `
   <withTheme(Connect(TestnetBadgeComponent)) />
   <ScrollViewMock showsVerticalScrollIndicator={false} contentContainerStyle={{...}}>
     <View style={{...}}>
-      <withTheme(AccountAddressComponent) account={{...}} token={{...}} />
+      <withTheme(AccountAddressComponent) account={{...}} token={[undefined]} />
```
krisboit: is it ok to be undefined ?

PR #384, opened 2020-02-24
Cosmos sign

georgeburduhos:


src/core/utils/http-client.ts
```diff
@@ -0,0 +1,51 @@
...
+            body: JSON.stringify({
+                jsonrpc: '2.0',
+                id,
+                method,
+                params: Array.isArray(params) ? params : [params]
+            })
+        };
+
+        // console.log('RPC client', 'request', request);
+        return fetch(this.url, request).then(async res => {
```
krisboit: you should use `post` method from this class

src/core/utils/http-client.ts
```diff
@@ -0,0 +1,51 @@
...
+            headers: {
+                'Content-Type': 'application/json'
+            },
+            body: JSON.stringify(body)
+        }).then(response => {
+            return response.json();
+        });
+    }
+
+    public rpcCall(method: string, params: any[] = []): Promise<any> {
```
krisboit: we should rename this to `jsonRpc()`

src/core/blockchain/cosmos/transaction.ts
```diff
@@ -116,12 +113,12 @@ export const buildTransferTransaction = async (
...
         feeOptions: tx.feeOptions,
         broadcatedOnBlock: undefined,
-        nonce: 0,
+        nonce: accountInfo.sequence,
         status: TransactionStatus.PENDING,
         additionalInfo: {
             account_number: accountInfo.account_number,
             memo: tx.extraFields.memo,
-            sequence: accountInfo.sequence
+            stdSignMsg
```
krisboit: you could use data.raw for this. like on zilliqa

PR #387, opened 2020-02-25
Filter transactions list on token page

NovacAndrei:


src/redux/wallets/selectors.ts
```diff
@@ -54,6 +56,31 @@ export const getAccountTransactions = (
     }
 };
 
+export const getAccountFilteredTransactions = (
+    state: IReduxState,
+    accountIndex: number,
+    blockchain: Blockchain,
+    token: ITokenConfig
+): IBlockchainTransaction[] => {
+    const account: IAccountState = getSelectedWallet(state).accounts.find(
```
krisboit: store `getSelectedWallet(state)` in a const

src/redux/wallets/selectors.ts
```diff
@@ -54,6 +56,31 @@ export const getAccountTransactions = (
...
+        acc => acc.index === accountIndex && acc.blockchain === blockchain
+    );
+    const transactions = getSelectedWallet(state).transactions;
+    if (transactions) {
+        return Object.values(getSelectedWallet(state).transactions)
+            .filter(
+                tx =>
+                    tx.address === account.address &&
```
krisboit: you should check the blockchain also

src/components/keyboard-custom/keyboard-custom.tsx
```diff
@@ -19,7 +19,7 @@ export interface IProps {
     showNumeric?: boolean;
     handleTextUpdate: (key: any) => void;
     handleDeleteKey: () => void;
-    buttons?: Array<{ label: string; onPress: () => void; style?: {}; disabled?: boolean }>;
+    buttons?: { label: string; onPress: () => void; style?: {}; disabled?: boolean }[];
```
krisboit: something is weird here :))) @maricalucian is adding Array you are deleting it :))

NovacAndrei: :))) I think it's because we have different typescript versions (maybe the global ones). I have seen then when I commit & push from Github Desktop, the code automatically changes from `Array` to `[ ]` :)) 

I have
```
tsc --version
Version 3.7.5
```

PR #415, opened 2020-02-28
Fix extension build

NovacAndrei:
- Crashed on lib `react-native-fast-image` because it was using `requireNativeComponent`


src/core/utils/fast-image.web.ts
```diff
@@ -0,0 +1,7 @@
+import { Image } from 'react-native';
+
+export default class FastImage extends Image {
+    static preload(sources: any): void {
```
krisboit: if this is not implemented and not used, delete it.

PR #417, opened 2020-02-28
Extension fix loading modal

NovacAndrei:
- Also fix Onboarding Screen Web design

### DIFF
--- a/src/components/loading-modal/loading-modal.tsx
+++ b/src/components/loading-modal/loading-modal.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { View, ActivityIndicator, Modal } from 'react-native';
+import { View, ActivityIndicator, Modal, Platform } from 'react-native';
 import { withTheme, IThemeProps } from '../../core/theme/with-theme';
 import stylesProvider from './styles';
 import { smartConnect } from '../../core/utils/smart-connect';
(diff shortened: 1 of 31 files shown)


src/components/loading-modal/loading-modal.tsx
```diff
@@ -23,6 +23,21 @@ const mapStateToProps = (state: IReduxState) => {
 export const LoadingModalComponent = (
     props: IReduxProps & IExternalProps & IThemeProps<ReturnType<typeof stylesProvider>>
 ) => {
+    if (Platform.OS === 'web') {
```
krisboit: so `Modal` component from `react-native` is not working well on web ?

NovacAndrei: image

`Modal` it's not working on web

NovacAndrei: I made a little research, it seems that `modal` it's not supported yet on web Platform.

But instead this could be a quick fix if we want to use Modal on web (but I have not tested it, yet):
https://github.com/Dekoruma/react-native-web-modal

krisboit: yeah true, i think it's better :)
we can use https://github.com/Dekoruma/react-native-web-modal/tree/master/packages/modal-enhanced-react-native-web from there.

PR #421, opened 2020-02-28
Near Recover Flow

NovacAndrei:


krisboit:
Let's postpone this a bit, almost all the code from these screens is Near specific, we have to define a solid structure for blockchain specific components. On Cosmos we will have some specific components also...

PR #440, opened 2020-03-02
config token screen

georgeburduhos:


src/screens/token/components/delegation-token/delegation-token.tsx
```diff
@@ -0,0 +1,91 @@
+import React from 'react';
```
krisboit: do we need this component ?

src/screens/token/token.tsx
```diff
@@ -205,68 +132,34 @@ export class TokenScreenComponent extends React.Component<
...
-                            onDonePressed={this.openSettingsMenu}
-                            account={this.props.account}
-                            wallet={this.props.wallet}
-                            chainId={this.props.chainId}
-                        />
-                    )}
-                </ScrollView>
-
-                <PasswordModal obRef={ref => (this.passwordModal = ref)} />
+                {token.ui.tokenScreenComponent === TokenScreenComponentType.DEFAULT ? (
```
krisboit: use a switch here, and make default value TokenScreenComponentType.DEFAULT.

src/screens/token/components/delegation-token/delegation-token.tsx
georgeburduhos: Its the second tab

PR #441, opened 2020-03-03
show content in extension while waiting for phone connection

maricalucian:


src/screens/token/token.tsx
```diff
@@ -60,7 +61,8 @@ export const mapStateToProps = (state: IReduxState, ownProps: INavigationParams)
         ),
         wallet: getSelectedWallet(state),
         extensionTransactionPayload: ownProps.extensionTransactionPayload,
-        chainId: getChainId(state, ownProps.blockchain)
+        chainId: getChainId(state, ownProps.blockchain),
+        canSend: Platform.OS !== 'web' || state.ui.extension.stateLoaded
```
krisboit: it's ok like this. As a hint for future, it would be more easier to understand this form:
```ts
canSend: Platform.select({
  default: true,
  web: state.ui.extension.stateLoaded
})
```

PR #462, opened 2020-03-10
modify state sent to extension

maricalucian:
don't send selected account, selected blockchain, selected wallet
send user preferences
added device id to state


krisboit:
@maricalucian you can merge this one after resolving conflicts

PR #473, opened 2020-03-13
Password Modal - Operation could not be completed

NovacAndrei:
Also:
- Center title on Create and Recover wallet screens
- Remove border bottom from our custom TextInput used in CustomKeyboard


ios/Podfile.lock
```diff
@@ -298,6 +298,8 @@ PODS:
   - React-jsinspector (0.61.1)
   - react-native-aes (1.3.6):
     - React
+  - react-native-background-timer (2.2.0):
```
georgeburduhos: why is this a change?

NovacAndrei: It was added on `pod install`

PR #486, opened 2020-03-18
add support url to settings

georgeburduhos:


src/screens/settings/settings.tsx
```diff
@@ -20,6 +28,8 @@ import { changePIN } from '../../redux/wallets/actions';
 import { isFeatureActive, RemoteFeature } from '../../core/utils/remote-feature-config';
 import { DebugModal } from '../../components/debug-modal/debug-modal';
 
+export const SUPPORT_URL = 'https://moonlet.xyz/links/support';
```
krisboit: do a rebase and add this url in configs :)

src/config/config-interface.ts
```diff
@@ -1,4 +1,5 @@
 export interface IAppConfig {
+    supportUrl: string;
     example: string;
```
krisboit: let's remove dummy vars :) i left them there just as an example

PR #487, opened 2020-03-18
Send Flow - fees validation

NovacAndrei:


src/core/blockchain/types/client.ts
```diff
@@ -36,6 +37,7 @@ export abstract class BlockchainGenericClient {
     public abstract calculateFees(
         from: string,
         to: string,
+        tokenType: TokenType,
```
krisboit: move this at the end, and leave it optional, this way we could use this method to estimate costs on all our smart contract calls not only token transfers.

also give a default value:
```
tokenType: TokenType = TokenType.NATIVE
```

src/screens/send/components/gas-fee-advanced/gas-fee-advanced.tsx
```diff
@@ -10,19 +10,22 @@ import { Blockchain } from '../../../../core/blockchain/types';
...
     blockchain: Blockchain;
     onInputFees: (gasPrice: string, gasLimit: string, feeTotal: string) => any;
+    insufficientFundsFees: boolean;
 }
 interface IState {
     inputGasPrice: string;
     inputGasLimit: string;
     displayErrorGasPrice: boolean;
     displayErrorGasLimit: boolean;
+    gasPriceUnit: string;
```
krisboit: why do we need this in state ?

PR #512, opened 2020-03-22
some extension bugs fixes

maricalucian:


src/components/testnet-badge/testnet-badge.tsx
```diff
@@ -23,7 +23,7 @@ const mapStateToProps = (state: IReduxState) => {
 
     return {
         blockchain,
-        networkName: blockchain && getNetworkName(state, blockchain),
+        networkName: blockchain && state.preferences && getNetworkName(state, blockchain),
```
krisboit: these checks (`blockchain && state.preferences`) should be moved inside `getNetworkName()`, otherwise we will have to copy them everywhere

maricalucian: true. moved the checks inside the selector

PR #518, opened 2020-03-23
update firebase setDefaults

georgeburduhos:


src/core/utils/remote-feature-config.ts
```diff
@@ -21,6 +21,14 @@ export const getRemoteConfigFeatures = async () => {
         duration = 0;
     }
 
+    // Set default values
+    firebase.config().setDefaults({
+        [RemoteFeature.NEAR]: false,
```
krisboit: these defaults are not quite ok :)
i think, they should be an empty array as string, like: `'[]'`

PR #524, opened 2020-03-25
Show Hints - Reveal hidden features/buttons automatically

NovacAndrei:


src/screens/transactions-history/list-transactions-history/list-transactions-history.tsx
```diff
@@ -74,7 +74,19 @@ export class TransactionsHistoryListComponent extends React.Component<
...
+                        backgroundColor:
+                            tx.status === TransactionStatus.PENDING
+                                ? theme.colors.cardBackground
+                                : theme.colors.appBackground,
+                        paddingVertical:
+                            tx.status === TransactionStatus.PENDING
+                                ? BASE_DIMENSION * 2
+                                : BASE_DIMENSION
+                    }
+                ]}
```
georgeburduhos: you can move this in styles
stylePending { backgroundColor: theme.colors.cardBackground, paddingVertical: BASE_DIMENSION*2}
styleOther { backgroundColor: appBackground, paddingVertical: BASE_DIMENSION}

src/redux/app/actions.ts
```diff
@@ -1,11 +1,18 @@
...
+export const SHOW_HINT = 'SHOW_HINT';
 
 export const appSetAcceptedTcVersion = (version: number) => {
     return {
         type: SET_ACCEPTED_TC_VERSION,
         data: version
     };
 };
+
+export const showHint = (screen: HintsScreen, component: HintsComponent) => {
```
krisboit: `showHint` is a little misleading in this case because when calling this action it doesn't show anything. Maybe `updateDisplayedHints`, i'm also ok if you find a better name :)

src/screens/manage-account/manage-account.tsx
```diff
@@ -71,21 +80,39 @@ export class ManageAccountComponent extends React.Component<
...
+    private showHints() {
+        const tokens = Object.keys(this.props.selectedAccount.tokens);
+        if (
+            tokens &&
+            tokens[1] &&
+            this.props.hints.MANAGE_ACCOUNT.TOKENS_LIST < DISPLAY_HINTS_TIMES
+        ) {
+            const id = tokens[1];
+
+            setTimeout(() => {
```
krisboit: we may want to trigger hints display from other places, i would recommend to not use `setTimeout` here. Use it in `componentDidMount`

krisboit: it's applicable for all occurrences :)

PR #538, opened 2020-03-27
Legal Modal - Show back button on Create / Recover Flow

NovacAndrei:


src/components/legal/legal-modal/legal-modal.tsx
```diff
@@ -28,6 +28,7 @@ const mapStateToProps = (state: IReduxState) => {
 interface IState {
     showModal: boolean;
     tcLatestVersion: number;
+    isOnboardingFlow: boolean;
```
georgeburduhos: make it more generic like: allowBackAction or somthing similar

PR #542, opened 2020-03-29
Refactor tokens

georgeburduhos:


src/core/blockchain/types/config.ts
```diff
@@ -11,9 +12,12 @@ export interface IBlockchainConfig {
     derivationType: DerivationType;
     coin: string;
     defaultUnit: string;
+    defaultChainId: string;
     iconComponent: React.ComponentType<any>;
     tokens: {
-        [symbol: string]: ITokenConfig;
+        [chainId: string]: {
```
krisboit: this overcomplicates everything at core/blockchain level. I'm afraid we will have to revert this :( 
So let's keep here only native tokens, that will be the same on all blockchains. With a new PR we will handle the list of default tokens list by adding them in redux token config.

src/core/blockchain/types/config.ts
```diff
@@ -11,9 +12,12 @@ export interface IBlockchainConfig {
     derivationType: DerivationType;
     coin: string;
     defaultUnit: string;
+    defaultChainId: string;
```
krisboit: this shouldn't be here :)

krisboit: it will introduce a lot of problems...

src/redux/tokens/state.ts
```diff
@@ -0,0 +1,44 @@
...
+    };
+    units?: {
+        [unit: string]: BigNumber;
+    };
+    symbolMap?: {
+        [testnet: string]: string;
+    };
+}
+
+export interface ITokenState {
```
krisboit: since this is used at account level only , move it in wallets state

src/redux/tokens/actions.ts
```diff
@@ -0,0 +1,23 @@
...
+    dispatch: Dispatch<any>,
+    getState: () => IReduxState
+) => {
+    const blockchain = account.blockchain;
+    const chainId = getChainId(getState(), blockchain).toString();
+    dispatch({
+        type: ADD_TOKEN,
+        data: { token, chainId, blockchain }
+    });
+    addTokenToAccount(account, convertTokenConfig(token, account))(dispatch, getState);
```
krisboit: `convertTokenConfig` is returning a default order of `0`, when adding a token we should add it at the end of the list, so you can use `order: 999`

src/redux/tokens/static-selectors.ts
```diff
@@ -0,0 +1,78 @@
...
+            };
+            tokenValue[symbolKey] = accountToken;
+        });
+        tokenList[chainIdKey] = tokenValue;
+    });
+
+    return tokenList;
+};
+
+export const convertTokenConfig = (
```
krisboit: we should rename this to `generateAccountTokenState()`. It would be much more clear.

src/redux/tokens/actions.ts
georgeburduhos: convertTokenConfig(generateAccountTokenState) has a optional account, if the account is given then the order returned is the number of tokens + 1 so each time a new token is added it will be the last

PR #550, opened 2020-03-30
dropped transactions

georgeburduhos:


src/redux/wallets/reducer.ts
```diff
@@ -154,6 +155,23 @@ export default (state: IWalletsState = intialState, action: IAction) => {
                 }
             };
 
+        case UPDATE_TRANSACTION_STATUS:
```
maricalucian: You could use `TRANSACTION_UPSERT` for this. Basically it does the same thing, but alsa adds as a new transaction is the transaction does not exists. Or do you think unwanted errorenous insert could be an issue?

georgeburduhos: I wanted to keep separate the transaction updates that come from push notifications and the ones that are based on app calls. This case will update only the status for a specific transaction. @krisboit ?

src/core/blockchain/cosmos/config.ts
```diff
@@ -32,6 +32,7 @@ export const config: IBlockchainConfig = {
     coin: 'ATOM',
     defaultUnit: 'UATOM',
     iconComponent: CosmosIcon,
+    nrBlocksPassed: 10,
```
krisboit: we need a more intuitive name: something like `droppedTxBlocksThreshold`

src/core/blockchain/ethereum/config.ts
```diff
@@ -30,6 +30,7 @@ export const config: IBlockchainConfig = {
     coin: 'ETH',
     defaultUnit: 'WEI',
     iconComponent: EthIcon,
+    nrBlocksPassed: 10,
```
krisboit: we need to do some research here, i remember on ethereum we can consider a transaction dropped after 50 blocks

src/redux/wallets/reducer.ts
krisboit: if the actions are doing the same thing, let's reuse them.

src/core/blockchain/ethereum/config.ts
```diff
@@ -30,6 +30,7 @@ export const config: IBlockchainConfig = {
     coin: 'ETH',
     defaultUnit: 'WEI',
     iconComponent: EthIcon,
+    droppedTxBlocksThreshold: 10,
```
krisboit: for ethereum let's use 50

Issue #546, opened 2020-03-29
Ledger Wallet blockchain bottom navigation is visible


krisboit:
For a ledger wallet, bottom blockchain navigation should not be visible

![Screenshot_20200329-222134.png](https://images.zenhubusercontent.com/5cae495acda34f221bec9559/55e8bb9a-e099-49eb-886e-a34efdc2cafd)

![Screenshot_20200329-222141.png](https://images.zenhubusercontent.com/5cae495acda34f221bec9559/a61c544a-1b12-4775-8b94-81f0aad773a2)

PR #551, opened 2020-03-30
#546 - Hide blockchain bottom navigation for ledger wallets

georgeburduhos:


src/screens/dashboard/dashboard.tsx
```diff
@@ -395,6 +396,11 @@ export class DashboardScreenComponent extends React.Component<
         const showCreateAccount =
             this.props.isCreateAccount && this.props.selectedBlockchainAccounts?.length === 0;
 
+        /* Hardware wallets can have only one blockchain active */
+        let renderBottomNav = false;
+        const isHWWallet = this.props.wallet ? this.props.wallet.type === WalletType.HW : false;
```
krisboit: Hint: in future you can use something like this:
```
const isHWWallet = this.props?.wallet?.type === WalletType.HW;
```

PR #556, opened 2020-03-31
Add token notification support

maricalucian:


src/core/blockchain/ethereum/client.ts
```diff
@@ -13,11 +13,13 @@ import { config } from './config';
 import { convertUnit } from '../ethereum/account';
 import abi from 'ethereumjs-abi';
 import { Erc20Client } from './tokens/erc20-client';
-import { TokenType } from '../types/token';
+import { TokenType, TokenScreenComponentType } from '../types/token';
 import { NameService } from './name-service';
 import { getTransactionStatusByCode } from './transaction';
 import { generateAccountTokenState } from '../../../redux/tokens/static-selectors';
 
+const inputRegex = /^(\w{10})(\w{24})(\w{40})(\w{64})$/;
```
krisboit: instead of regex, use `decodeInputData` function:
```ts
const decodeInputData = (signature, data) => {
  const sig = signature.split(":");
  const sigParts = sig[0].split("(");
  const methodName = sigParts[0];
  let params = [];
  if (sigParts[1]) {
    params = sigParts[1]
      .replace(")", "")
      .split(",")
      .map(p => p.trim());
  }

  const methodId = abi.methodID(methodName, params).toString("hex");

  data = data.replace("0x", "");

  if (data.indexOf(methodId) === 0) {
    rawParamsData = data.substr(methodId.length);
    return abi.rawDecode(params, Buffer.from(rawParamsData, "hex"));
  } else {
    throw new Error("Cannot decode data");
  }
};
```

src/core/blockchain/ethereum/client.ts
```diff
@@ -160,7 +162,7 @@ export class Client extends BlockchainGenericClient {
             this.http.jsonRpc('eth_getTransactionReceipt', transactionHash)
         ];
 
-        return Promise.all(rpcCalls).then(res => {
+        return Promise.all(rpcCalls).then(async res => {
```
krisboit: all method from clients should be kept as light as we can, it should only facilitate network requests. For data transformation create a function in `core/${blockchain}/transaction.ts` for this particular case: `buildTransactionFromBlockchainData(blockchainData): IBlockchainTransaction<...>`

src/core/blockchain/zilliqa/client.ts
```diff
@@ -131,7 +131,11 @@ export class Client extends BlockchainGenericClient {
                 .toLowerCase(),
             field,
             subFields
-        ]).then(response => response?.result);
+        ])
+            .then(response => response?.result)
```
krisboit: why is this needed, `call` method should do this already ...

src/core/blockchain/zilliqa/client.ts
```diff
@@ -162,21 +165,68 @@ export class Client extends BlockchainGenericClient {
...
-
             let fromAddress = getAddressFromPublicKey(txData.senderPubKey.replace('0x', ''));
-
             if (!isBech32(fromAddress)) {
                 fromAddress = toBech32Address(fromAddress);
             }
 
-            const txStatus = txData.receipt
-                ? txData.receipt.success
+            let toAddress = !isBech32(txData.toAddr)
```
krisboit: same for this, let's move it to transaction.ts

src/core/blockchain/zilliqa/tokens/zrc2-client.ts
```diff
@@ -34,22 +34,34 @@ export class Zrc2Client {
     }
 
     public async getTokenInfo(contractAddress) {
-        const smartContractSubState = await this.client.getSmartContractSubState(
-            contractAddress,
-            'implementation'
-        );
+        const smartContractSubState = await this.client
+            .getSmartContractSubState(contractAddress, 'implementation')
+            .catch(e => {
```
krisboit: this is ok to fail, the pattern that we use everywhere is to wrap the code in `try...catch`

maricalucian: Yup, moved catch to proper  location.

src/core/blockchain/ethereum/client.ts
maricalucian: done

src/core/blockchain/ethereum/client.ts
maricalucian: refactore and moved this in client-utils.ts

src/core/blockchain/ethereum/client-utils.ts
```diff
@@ -0,0 +1,131 @@
...
+            data,
+            feeOptions: {
+                gasPrice: txInfo.gasPrice,
+                gasLimit: txInfo.gas,
+                feeTotal: txReceipt.gasUsed
+            },
+            broadcatedOnBlock: txInfo.blockNumber,
+            nonce: txInfo.nonce,
+            status: getTransactionStatusByCode(txReceipt.status),
+            token: generateAccountTokenState(token)
```
krisboit, line 85: @georgeburduhos we should use here the token config not the token state from account...
@maricalucian we can leave it like this for now

src/core/blockchain/types/client.ts
```diff
@@ -15,6 +16,7 @@ export abstract class BlockchainGenericClient {
     public nameService: IBlockchainNameService;
     public http: HttpClient;
     public readonly chainId: ChainIdType;
+    public clientUtils: IClientUtils;
```
krisboit, line 19: the property name should be `utils` it would be more simple when using it `getClient(...).utils....`

src/core/blockchain/zilliqa/client-utils.ts
```diff
@@ -0,0 +1,126 @@
...
+import { TransactionStatus } from '../../wallet/types';
+import { generateAccountTokenState } from '../../../redux/tokens/static-selectors';
+
+export class ClientUtils implements IClientUtils {
+    constructor(private client: Client) {}
+
+    async getTransaction(hash: string): Promise<IBlockchainTransaction> {
+        const txData = await this.client.http.jsonRpc('GetTransaction', [hash]).then(response => {
+            if (!response.result) {
+                throw new Error(
```
krisboit, line 18: when returning a promise it's better to return a `Promise.reject()` than throwing an exception.

src/redux/wallets/actions.ts
```diff
@@ -473,10 +477,8 @@ export const updateTransactionFromBlockchain = (
...
                 accountIndex: transactionAccount.index,
-                token: transactionAccount.tokens[getBlockchain(blockchain).config.coin],
-                tokenLogo: getBlockchain(blockchain).config.tokens[
-                    getBlockchain(transactionAccount.blockchain).config.coin
-                ].icon
+                token: generateAccountTokenState(tokenConfig),
+                tokenLogo: tokenConfig.icon
             };
 
             dispatch(setSelectedWallet(wallet.id));
```
krisboit, line 484: it would be good to trigger a balance update on the addresses affected

src/core/blockchain/ethereum/client-utils.ts
georgeburduhos: I'll make the changes for transactions in a separate PR

Issue #547, opened 2020-03-29
Ledger BLE Send page messages


krisboit:
When signing a transaction using ledger, if the ledger device is not opened, the message displayed is `Signing transaction`. The message should be: `Connecting to your ledger. Make sure your device is unlocked.`

Issue #544, opened 2020-03-29
Ledger BLE Connect flow messages


krisboit:
When connecting a ledger to moonlet:

1. After user successfully pairs the ledger, we display a loader and no message. We should display `Open ${blockchain} app on you ledger to continue...`

PR #559, opened 2020-03-31
#544 #547 ledger messages

georgeburduhos:


src/redux/wallets/actions.ts
```diff
@@ -514,13 +518,6 @@ export const sendTransferTransaction = (
         });
 
         if (appWallet.type === WalletType.HW) {
-            dispatch({
```
krisboit: why did you remove this one ?

I think it would be good to display these messages:
1. connecting to ledger
2. make sure app is opened
3. review tx
4. broadcasting

PR #562, opened 2020-03-31
Application doesn't start in airplane mode

NovacAndrei:
- fixed Android
- iOS was already working


src/core/utils/remote-feature-config.ts
```diff
@@ -29,22 +29,27 @@ export const getRemoteConfigFeatures = async () => {
...
+                RemoteFeature.TC_VERSION
+            ]);
+
+        featuresConfig = {};
+        // Retrieve values
+        Object.keys(objects).forEach(key => {
+            featuresConfig[key] = objects[key].val();
+        });
+    } catch (err) {
+        //
```
krisboit: i think we need to handle this, featuresConfig will be undefined, we should set some defaults

PR #566, opened 2020-04-01
 Firebase set fetch duration

NovacAndrei:


src/config/config-beta.ts
```diff
@@ -3,7 +3,8 @@ import { IAppConfig } from './config-interface';
 export const CONFIG: IAppConfig = {
     supportUrl: 'https://moonlet.xyz/links/support',
     env: process.env.MOONLET_SOME_KEY,
-    tokensUrl: 'https://static.moonlet.dev/tokens/'
+    tokensUrl: 'https://static.moonlet.dev/tokens/',
+    firebaseFetchDuration: 0
```
krisboit, line 7: a better name would be: `firebaseConfigFetchInterval`

PR #568, opened 2020-04-01
Fix Create Password Flow

NovacAndrei:
- on create/recover/connect flow, if you already have a PIN setup, don't create another one, request the current PIN
- also clear keychain storage when the last wallet has been deleted
- fix crashing bottom sheet on connect flow

### DIFF
--- a/src/components/password-modal/password-modal-component.tsx
+++ b/src/components/password-modal/password-modal-component.tsx
@@ -118,9 +118,13 @@ export class PasswordModalComponent extends React.Component<
         clearInterval(this.countdownListener);
     }
 
-    public static async getPassword(title?: string, subtitle?: string) {
+    public static async getPassword(
+        title?: string,
+        subtitle?: string,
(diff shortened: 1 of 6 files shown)


src/components/password-modal/password-modal-component.tsx
```diff
@@ -152,19 +152,32 @@ export class PasswordModalComponent extends React.Component<
...
-            title: translate('Password.setupPinTitle'),
-            subtitle: subtitle || translate('Password.createPinSubtitle'),
-            currentStep: ScreenStep.CREATE_PIN_TERMS,
-            enableBiometryAuth: false,
-            allowBackButton: true,
-            showAttempts: false
-        });
+        const passwordCredentials = await getPassword();
+        if (passwordCredentials && passwordCredentials.password) {
+            this.setState({
```
krisboit, line 161: this doesn't seem to be right :)
create password should create password, we shouldn't start this flow if a password is already present, or much better this flow should return a rejection is there is a password already set

PR #582, opened 2020-04-02
 Reveal Private Key Modal - is not displayed

NovacAndrei:


src/screens/token/components/account-settings/account-settings.tsx
```diff
@@ -49,33 +49,43 @@ export class AccountSettingsComponent extends React.Component<IProps & IExternal
...
+            this.setState({ isModalVisible: this.props.isVisible });
+        }
+    }
+
+    public revealPrivateKey() {
         try {
-            // TODO: not working because of modal over modal
-            const password = await PasswordModal.getPassword();
+            this.setState({ isModalVisible: false }, async () => {
+                const password = await PasswordModal.getPassword();
```
krisboit, line 66: i think we could have the same problems here, as at this moment the account options modal is not completely closed. A safe method would be to open password modal when onModalHide is called on account options modal.

src/screens/token/components/account-settings/account-settings.tsx
```diff
@@ -49,33 +49,43 @@ export class AccountSettingsComponent extends React.Component<IProps & IExternal
...
             showSecurityWarning: false,
-            isLoading: false
+            isLoading: false,
+            isModalVisible: props.isVisible
         };
     }
 
-    public async revealPrivateKey() {
+    public componentDidUpdate(prevProps: IExternalProps) {
+        if (this.props.isVisible !== prevProps.isVisible) {
```
krisboit, line 58: we can avoid using lifecycle function in this case. 
- keep the visibility state of account options modal in props
- in state add the visibility state of other modal
Example:

```ts
interface IState {
  ...
  displayOtherModal: boolean;
}
...
render() {
...
  <Modal isVisible={this.props.visible && !this.state.displayOtherModal}>
...
}
```

PR #588, opened 2020-04-03
 Handle current date for Wrong Password

NovacAndrei:


src/components/password-modal/password-modal-component.tsx
```diff
@@ -244,15 +309,18 @@ export class PasswordModalComponent extends React.Component<
...
     private async handleWrongPassword() {
-        this.setCountdownListener();
-
         if (this.state.showAttempts) {
             this.props.incrementFailedLogins();
             this.handlePasswordAttempts();
 
             const failedLoginBlocking = FAILED_LOGIN_BLOCKING[this.props.failedLogins];
             if (failedLoginBlocking) {
                 this.props.setAppBlockUntil(new Date(new Date().getTime() + failedLoginBlocking));
```
krisboit, line 318: this still uses device time, so it can be hacked

NovacAndrei: I left here a TODO because we have to think about a scenario.
I am going to think and come with a solution, but also if you have some recommendations 🙂

krisboit: yeah, i think we should block the app till the user connects to internet. After we get the current time we start the countdown.

PR #606, opened 2020-04-07
add security warning for unsafe environments

krisboit:


src/components/security-checks/styles.ts
```diff
@@ -0,0 +1,48 @@
...
+            lineHeight: normalize(25),
+            marginTop: BASE_DIMENSION * 6
+        },
+        imageContainerStyle: {
+            flex: 1,
+            alignItems: 'center',
+            justifyContent: 'center'
+        },
+        imageStyle: {
+            width: Dimensions.get('window').width * 0.5,
```
NovacAndrei, line 44: We also have `pw` and `ph` for this
image

src/components/security-checks/security-checks.tsx
```diff
@@ -0,0 +1,136 @@
...
+                                // disabled={!acknowledged}
+                                onPress={() => this.continue()}
+                            >
+                                {translate('App.labels.continue')}
+                            </Button>
+                        </View>
+                    </View>
+                </SafeAreaView>
+
+                {/* <Button onPress={() => BackHandler.exitApp()}>EXIT</Button>
```
NovacAndrei, line 127: If not needed, maybe this can be removed :)

src/components/security-checks/styles.ts
krisboit: uh nice! :) i didn't knew about them

PR #620, opened 2020-04-09
Extension Connexion Management Screen

NovacAndrei:


src/screens/connect-extension/connect-extension.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+    constructor(props: IThemeProps<ReturnType<typeof stylesProvider>>) {
+        super(props);
+        this.state = {
+            isConnected: WalletConnectClient.isConnected()
+        };
+    }
+
+    public componentDidMount() {
+        this.connectionInterval = setInterval(() => {
```
krisboit, line 42: this is weird, basically we verify every 500ms if we have connection, and if we don't we delete connection details.
There may be cases when the connection could take longer

krisboit: i think we should not delete connection

src/screens/connect-extension/connect-extension.tsx
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    private async disconnectExtension() {
+        if (
+            await DialogComponent.confirm(
+                translate('ConnectExtension.disconnect'),
+                translate('ConnectExtension.disconnectInfo')
+            )
+        ) {
+            WalletConnectClient.disconnect();
```
krisboit, line 69: here we should delete connection info

src/screens/connect-extension/connect-extension.tsx
krisboit: also maybe Wallet connect has some events for connect and disconnect events, this way we could refactor this interval

NovacAndrei: I made some changes and improvements, but I left a TODO with one more issue to solve. I'll continue to search for a fix.

src/core/wallet-connect/wallet-connect-client.ts
```diff
@@ -146,6 +146,13 @@ export const WalletConnectClient = (() => {
         });
     };
 
+    const onDisconnect = (callback: (payload: any) => void) => {
+        walletConnector &&
+            walletConnector.on('disconnect', (error, payload) => {
```
krisboit, line 151: you can do something similar to detect connection.
```
walletConnector.on('connect', ...);
```

PR #621, opened 2020-04-09
Hold to unveil sensitive data

NovacAndrei:
- Create
- Recover
- Unveil (Wallets)
- Private key


src/screens/create-wallet-mnemonic/create-wallet-mnemonic.tsx
```diff
@@ -86,9 +88,21 @@ export class CreateWalletMnemonicScreenComponent extends React.Component<
...
+                                    ? this.state.mnemonic[indexFrom + i]
+                                    : '**********'
+                            }`}
                         </Text>
                     ))}
+
+                    <Button
+                        style={styles.unveilButton}
+                        onPressIn={() => this.setState({ unveilMnemonic: true })}
+                        onPressOut={() => this.setState({ unveilMnemonic: false })}
```
krisboit, line 102: for `onPressOut` let's use a setTimeout to delay a little the hidden state. we can use 2-300 ms as delay. This way if the user taps briefly on unveil button it will be more clear what's happening.

src/screens/recover-wallet/recover-wallet.tsx
```diff
@@ -139,17 +146,15 @@ export class RecoverWalletScreenComponent extends React.Component<
...
-                onPress: () => this.pasteFromClipboard(),
-                disabled: this.mnemonicsFilled()
-            });
-        }
-
-        keyboardButtons.push(
+        const keyboardButtons: IKeyboardButton[] = [
+            {
+                label: translate('App.labels.holdUnveil'),
+                onPress: () => {
```
krisboit, line 152: this could be removed

PR #622, opened 2020-04-09
Fix Token - Manage Account

NovacAndrei:
**Problem:** On `Manage Account Modal`, if you go to background and return to the app, the `Manage Account Modal` does not disappear and  the `Password Modal` is not shown (we are not requesting the password anymore).

**Solution:** check for app state, if app state is `background` or `inactive`, set visible `false` for the `Manage Account Modal` and the `Password Modal` will be displayed.


krisboit:
i think we can close this one, right ?

PR #639, opened 2020-04-13
Update keychain.ts

georgeburduhos:
Fix create and recover


src/core/secure/keychain.ts
```diff
@@ -116,9 +118,6 @@ export const getPinCode = async () => {
     } catch (error) {
         // console.log('keychain error', JSON.stringify(error, null, 4));
         if (error.message.indexOf('Authentication failed') >= 0) {
-            if (DeviceInfo.getManufacturerSync() === 'OnePlus') {
```
krisboit, line 119: is this really intended ?

PR #658, opened 2020-04-14
Zrc fixes

georgeburduhos:


src/core/i18n/translation/locales/en.ts
```diff
@@ -376,7 +376,8 @@ export const translation: ITranslations = {
             TR_UNDERPRICED: 'Transaction underpriced: gas price might be too low',
             NOT_ENOUGH_TOKENS: 'Insufficient funds for transaction',
             COSMOS_ERROR_12: 'Insufficient fee',
-            COSMOS_ERROR_14: 'Insufficient funds for transaction'
+            COSMOS_ERROR_14: 'Insufficient funds for transaction',
+            CONTRACT_TX_NORMAL_NOT_ALLOWED: "Contract account won't accept normal txn"
```
krisboit, line 380: This text is not friendly with ordinary users. let's use: `${address} is a smart contract that doesn't accept normal transactions.`

krisboit: verify it with Adrian please.

src/screens/manage-token/manage-token.tsx
```diff
@@ -176,31 +176,39 @@ export class ManageTokenComponent extends React.Component<
...
+                    getBlockchain(blockchain)
+                        .getClient(this.props.chainId)
+                        .tokens[tokenType].getTokenInfo(this.state.fieldInput)
+                ]);
+                if (getTokenInfo[0].status === 200) {
+                    staticToken = await getTokenInfo[0].json();
+                }
+                blockchainToken = getTokenInfo[1];
+            } catch {
+                //
```
krisboit, line 194: We should handle errors, maybe display an error message for the user.

src/screens/manage-token/manage-token.tsx
```diff
@@ -176,31 +176,39 @@ export class ManageTokenComponent extends React.Component<
...
+                if (fetchResponse.status === 200) {
+                    staticToken = await fetchResponse.json();
+                    blockchainToken = await getBlockchain(blockchain)
+                        .getClient(this.props.chainId)
+                        .tokens[tokenType].getTokenInfo(staticToken.contractAddress);
+                } else {
+                    this.setState({ showError: true, isLoading: false });
+                }
+            } catch {
+                //
```
krisboit, line 211: We should handle errors, maybe display an error message for the user.

src/screens/manage-token/manage-token.tsx
georgeburduhos: Lower in the code we handle this. Since there are multiple versions of getting a token if non works we display an error message to user.

PR #661, opened 2020-04-14
Fix Android Multitasking background color

NovacAndrei:
On screens with `FLAG_SECURE` (sensitive data - when we forbid screenshots), when you go multitasking, a white screen was displayed.

Now it's change to our primary background color.


krisboit:
wow nice, i thought it was not possible :)

PR #662, opened 2020-04-14
Improve Send Screen Design

NovacAndrei:
- Send Screen more responsive
- add iPhone X support


src/screens/send/styles.ts
```diff
@@ -47,7 +54,11 @@ export default (theme: ITheme) =>
             backgroundColor: theme.colors.settingsDivider
         },
         bottomContainer: {
-            padding: BASE_DIMENSION + BASE_DIMENSION / 2,
+            padding: DEFAULT_BOTTOM_CONTAINER_PADDING,
+            paddingBottom: Platform.select({
+                default: DEFAULT_BOTTOM_CONTAINER_PADDING,
+                ios: isIphoneXorAbove() ? BASE_DIMENSION * 3 : DEFAULT_BOTTOM_CONTAINER_PADDING
```
krisboit, line 60: doe this fix our layout issue on smaller devices ? Looking at the code to me it seems only ios is impacted by these changes...

also we should solve this by using flex functionality.

NovacAndrei: I made some changes regarding the padding and margins. 
I have tested on both Pixel 2 and iPhone SE.

`Pixel 2`
image

`iPhone SE`
image

Also the scroll should work properly.
Please try again now with this code when you can.

krisboit: I think the problem appears on testnet, when the yellow badge is visible

NovacAndrei: @krisboit It seems to work ok on Pixel 2 with testnet badge:

https://drive.google.com/open?id=1rwlVpX3lFA-eMu9aMfXfr5gCPp1mzFDc

PR #667, opened 2020-04-15
On last wallet delete Reset PIN and biometric settings

NovacAndrei:


src/core/secure/keychain.ts
```diff
@@ -68,6 +66,16 @@ export const getBaseEncryptionKey = async () => {
     return password;
 };
 
+export const clearEncryptionKey = async () => {
+    try {
+        clearPinCode();
```
georgeburduhos, line 71: Don’t use clearPinCode here

src/redux/preferences/reducer.ts
```diff
@@ -44,20 +46,31 @@ export default (state: IPrefState = initialState, action: IAction): IPrefState =
...
             return {
                 ...state,
                 testNet: !state.testNet
             };
-        case TOGGLE_TOUCH_ID:
-            state = { ...state };
-            state.touchID = !state.touchID;
-            break;
+
+        case TOGGLE_BIOMETRIC_AUTH:
```
georgeburduhos, line 62: if you use toggle, why u made a new action for disable? toogle can make enable/disable

src/screens/settings/settings.tsx
```diff
@@ -128,7 +128,7 @@ export class SettingsScreenComponent extends React.Component<
                             <Switch
                                 onValueChange={async () => {
                                     try {
-                                        if (this.props.touchID) {
+                                        if (this.props.biometricActive) {
```
georgeburduhos, line 131: is this still necessary since when you delete all wallets this settings is reset to false?

src/screens/settings/settings.tsx
```diff
@@ -128,35 +128,29 @@ export class SettingsScreenComponent extends React.Component<
...
-                                                undefined,
-                                                {
-                                                    showCloseButton: true
-                                                }
-                                            );
-                                            await delay(0);
-                                            await setPinCode(password);
-                                        }
-                                        this.props.toggleTouchID();
+                                        this.props.openLoadingModal();
```
krisboit, line 131: i think you missed a flow here...
if biometric is active we have to disable it and remove pin code from keychain, no password modal.

PR #673, opened 2020-04-16
Fix code validation for undefined

NovacAndrei:


src/screens/send/components/fee-options/fee-options.tsx
```diff
@@ -70,18 +70,24 @@ export class FeeOptionsComponent extends React.Component<
...
+                    this.props.toAddress,
+                    1,
+                    tokenSendingToken.contractAddress,
+                    tokenSendingToken.type
+                );
+
+            this.setState({ feeOptions: fees });
+            this.props.onFeesChanged(fees);
+        } catch {
+            //
```
georgeburduhos, line 87: If the estimation fails we should at lease show something to user, live advanced fees or the default fees that are set in blockchain config

krisboit: yep, default fees would be ok

PR #675, opened 2020-04-16
update biometric

georgeburduhos:


krisboit:
It doesn't make sense to keep this opened, we have an issue opened for this, for not we will postpone- it

PR #699, opened 2020-04-27
Connect Extension - Mobile

NovacAndrei:


src/screens/connect-extension/connect-extension.tsx
```diff
@@ -29,42 +48,123 @@ export class ConnectExtensionScreenComponent extends React.Component<
...
+            connection.connectionId = res[1];
+            connection.encKey = extraData?.encKey;
+            connection.os = extraData?.os;
+            connection.platform = extraData?.browser;
+        }
+
+        return connection;
+    }
+
+    private async connectOrUpdateExtension(value: string) {
```
krisboit, line 108: connect and update should be separated, we can reuse code though

src/screens/connect-extension/connect-extension.tsx
```diff
@@ -74,59 +174,115 @@ export class ConnectExtensionScreenComponent extends React.Component<
...
+                    const request = {
+                        method: 'POST',
+                        headers: { 'Content-Type': 'application/json' },
+                        body: JSON.stringify({
+                            connectionId,
+                            authToken: sha256(authToken)
+                        })
+                    };
+
+                    try {
```
krisboit, line 196: you are already in a try..catch block, so this one is not needed

src/screens/connect-extension/connect-extension.tsx
```diff
@@ -74,59 +174,115 @@ export class ConnectExtensionScreenComponent extends React.Component<
...
+                        const disconnectResponse = await fetch(
```
krisboit, line 197: use http client instead

src/screens/connect-extension/connect-extension.tsx
```diff
@@ -74,59 +174,115 @@ export class ConnectExtensionScreenComponent extends React.Component<
...
+                            CONFIG.extSyncDisconnectUrl,
+                            request
+                        );
+
+                        const resData = await disconnectResponse.json();
+                        if (resData?.success === true) {
+                            // Delete connection from async storage
+                            await deleteFromStorage(CONN_EXTENSION);
```
krisboit, line 205: Delete from storage should happen even if disconnect call fails...

krisboit: if we have an error alt least we let the user to setup a new connection

src/screens/connect-extension/connect-extension.tsx
```diff
@@ -29,42 +48,132 @@ export class ConnectExtensionScreenComponent extends React.Component<
...
+            os: undefined,
+            platform: undefined
+        };
+
+        const valueSplit = value.split(qrCodeRegex);
+        if (valueSplit[1]) {
+            connection.connectionId = valueSplit[1];
+        }
+
+        let extraValueSplit = valueSplit[4] && valueSplit[4].split(qrCodeRegexExtraInfo);
```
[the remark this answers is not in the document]
NovacAndrei: `URLSearchParams` cannot be used on iOS

src/screens/connect-extension/connect-extension.tsx
```diff
@@ -29,42 +49,137 @@ export class ConnectExtensionScreenComponent extends React.Component<
...
+                    JSON.stringify(extensionState(store.getState())),
+                    connection.encKey
+                ),
+                authToken: sha256(connection.encKey),
+                fcmToken: await Notifications.getToken()
+            });
+
+            return res;
+        } catch (err) {
+            throw err;
```
krisboit, line 127: we shouldn't throw errors in async functions, it would be better to return a `Promise.reject()`

src/screens/connect-extension/connect-extension.tsx
```diff
@@ -29,42 +49,137 @@ export class ConnectExtensionScreenComponent extends React.Component<
...
+                await this.syncExtension(connectionParse);
+            }
+
+            this.setState({ isLoading: false });
+        } catch {
+            this.setState({ isLoading: false });
+        }
+    }
+
+    private async extractQrCodeData(value: string): Promise<IQRCode> {
```
krisboit, line 91: why is this `async` ?

src/screens/connect-extension/connect-extension.tsx
```diff
@@ -29,42 +49,137 @@ export class ConnectExtensionScreenComponent extends React.Component<
...
+                    }
+
                     this.setState({ isConnected: true });
-                } else {
-                    this.setState({ isConnected: false });
                 }
-            } catch (err) {
-                this.setState({ isConnected: false });
+            } catch {
+                //
```
krisboit, line 160: we should display an error in this case, a `Dialog.info()` would be fine.

src/core/utils/format-number.ts
```diff
@@ -50,3 +50,22 @@ export const formatNumber = (amount: number | BigNumber, options: INumberFormatO
 
 const integerRegex = new RegExp(/^\d+$/);
 export const isInteger = (value: string) => integerRegex.test(value);
+
+export const qrCodeRegex = /^mooonletExtSync:([^\@]*)\@([^\/\?]*)([^\?]*)?\??(.*)/;
```
krisboit, line 54: these stuff doesn't belong here :)
qrCode stuff can be moved where connection is done, it's specific to that feature

src/screens/connect-extension/connect-extension.tsx
```diff
@@ -74,59 +189,103 @@ export class ConnectExtensionScreenComponent extends React.Component<
...
+                    const authToken = connectionParse.encKey;
+
+                    const http = new HttpClient(CONFIG.extSyncDisconnectUrl);
+                    const res = await http.post('', {
+                        connectionId,
+                        authToken: sha256(authToken)
+                    });
+
+                    if (res?.success === true) {
+                        await this.deleteConnection();
```
krisboit, line 210: we should delete the connection even if the call to disconnect function fails

PR #708, opened 2020-05-07
Connect Extension - Send Transaction

NovacAndrei:


src/core/i18n/translation/locales/en.ts
```diff
@@ -412,6 +414,16 @@ export const translation: ITranslations = {
             currentlyActive: 'Currently active',
             qrCodeError: 'Invalid QR Code!',
             error: 'The connection has failed! Please try again!'
+        },
+        Notification: {
+            title: 'Confirm transaction',
```
krisboit, line 419: this is too generic, let's transform this into:
```ts
Notifications: {
  extensionTx: {
    title: '...'
    body: '...'
  }
}
```

src/redux/wallets/actions.ts
```diff
@@ -511,6 +513,66 @@ export const updateTransactionFromBlockchain = (
     }
 };
 
+export const signAndSendTransaction = (tx: any, password: string, requestId: string) => async (
```
krisboit, line 516: why don't we use `sendTransferTransaction` action creator ? Your implementation will not work at it should with ledger hw wallet...

krisboit: also all the code from here seems to be duplicated with code from `sendTransferTransaction`

PR #712, opened 2020-05-09
State migration

georgeburduhos:
@NovacAndrei @krisboit  - if all ok don’t approve both of you because we need to manually merge it without squash


NovacAndrei:
👍

PR #714, opened 2020-05-11
sentry

georgeburduhos:


src/core/secure/keychain.ts
```diff
@@ -4,12 +4,13 @@ import { storeEncrypted, readEncrypted, deleteFromStorage } from './storage';
...
 import uuidv4 from 'uuid/v4';
 import { Platform } from 'react-native';
+import CONFIG from '../../config/config-beta';
 
 const defaultOptions = {
-    serviceEncryption: 'com.moonlet.encryption',
-    usernameEncryption: 'moonlet-encryption-key',
-    usernamePin: 'moonlet-pin',
-    servicePin: 'com.moonlet.pin'
+    serviceEncryption: `${CONFIG.keychain}.encryption`,
```
krisboit, line 10: these are not backword compatible, right ?

krisboit: this is an issue only on iOS ? i remember i saw in plist that both moonlet release and moonlet beta are using sharing the same keystore... maybe it's safer to modify from there. Like we could leave production as it is now and change for beta

src/core/blockchain/zilliqa/client-utils.ts
```diff
@@ -14,8 +14,8 @@ export class ClientUtils implements IClientUtils {
     async getTransaction(hash: string): Promise<IBlockchainTransaction> {
         const txData = await this.client.http.jsonRpc('GetTransaction', [hash]).then(response => {
             if (!response.result) {
-                Promise.reject(
-                    response.error.message || `Error getting transaction info for ${hash}`
+                throw new Error(
```
krisboit, line 17: when we return promises promise reject should be the way to go...

PR #721, opened 2020-05-12
Refactor Bottom Component

NovacAndrei:


src/components/bottom-cta/bottom-cta.tsx
```diff
@@ -0,0 +1,137 @@
...
+import { Text, Button } from '../../library';
+import { smartConnect } from '../../core/utils/smart-connect';
+import { Amount } from '../amount/amount';
+import { normalize, BASE_DIMENSION } from '../../styles/dimensions';
+import { IAccountState } from '../../redux/wallets/state';
+import { ITokenConfigState } from '../../redux/tokens/state';
+import BigNumber from 'bignumber.js';
+
+export interface IExternalProps {
+    options: {
```
krisboit, line 14: i would make this component more dummy,

I think something like this would be more versatile:

```tsx
<BottomCta label="Do Magic" ctaStyle={{color: 'red'}} disabled={false} onPress={() => console.log('Let the magic begin!')}>
  <BottomCta.Row>{components here}</BottomCta.Row>
  <BottomCta.Row>{components here}</BottomCta.Row>
</BottomCta>
```

This way this component is very simple, without any application logic inside, it's just a dummy UI component.

PR #723, opened 2020-05-13
Delegate design

georgeburduhos:


src/core/blockchain/celo/config.ts
```diff
@@ -14,6 +14,39 @@ import {
     CELO_USD_TESTNET_BAKLAVA
 } from './tokens/cUSD';
 
+export const accountCTA = {
+    mainCta: {
+        title: 'App.labels.quickVote',
```
NovacAndrei, line 19: Why not use here `title: translate('App.labels.quickVote')`,
I think it's easier to `translate` the title in config, rather than everywhere it's used.

src/core/blockchain/cosmos/config.ts
```diff
@@ -15,7 +15,7 @@ export const ATOM_NATIVE: ITokenConfigState = {
     decimals: 6,
     ui: {
         decimals: 3,
-        tokenScreenComponent: TokenScreenComponentType.DEFAULT // TODO: DELEGATE is in progress
+        tokenScreenComponent: TokenScreenComponentType.DELEGATE // TODO: DELEGATE is in progress
```
NovacAndrei, line 18: You can remove the TODO

src/library/button/button.tsx
```diff
@@ -25,43 +26,61 @@ export interface IButtonProps {
...
-                    ? BASE_DIMENSION
-                    : BASE_DIMENSION + BASE_DIMENSION / 2
-            }
-        ]}
-    >
-        {props?.leftIcon && (
-            <Icon name={props.leftIcon} size={normalize(32)} style={props.styles.leftIcon} />
-        )}
-        <Text
+    <View style={[props.styles.container, props.wrapperStyle]}>
```
NovacAndrei, line 33: We have to make sure that all `Buttons` used in the App are fixed after this change

src/screens/token/components/delegate-token/delegate-token.tsx
```diff
@@ -87,7 +87,7 @@ export class DelegateTokenScreenComponent extends React.Component<
     //     extensionTransactionPayload: any; // TODO add typing
```
NovacAndrei, line 87: We can remove this if not needed

src/screens/validator-screen/validator-screen.tsx
```diff
@@ -0,0 +1,99 @@
...
+        IThemeProps<ReturnType<typeof stylesProvider>>
+> {
+    public static navigationOptions = navigationOptions;
+
+    public render() {
+        const { styles, blockchain, validator } = this.props;
+        const config = getBlockchain(blockchain).config;
+
+        const textTop = `${validator.totalLabel} (${validator.rank})`;
+        const amount = '2900000';
```
NovacAndrei, line 60: This should not be hardcoded

src/core/blockchain/celo/stats.ts
```diff
@@ -1,8 +1,227 @@
...
+        color: '#E91E63',
+        type: IStatValueType.AMOUNT,
+        data: {
+            value: '10000.00',
+            tokenSymbol: 'cGLD',
+            blockchain: Blockchain.CELO
+        }
+    },
+    {
+        title: 'Unlocking',
```
NovacAndrei, line 91: Maybe use here translate of keys from translate

georgeburduhos: Its dummy data for now, will adapt when the blockchain calls will be linked

src/core/blockchain/celo/stats.ts
```diff
@@ -1,8 +1,229 @@
...
+    IStatValueType,
+    AccountStats,
+    IValidatorCard,
+    CardActionType
+} from '../types/stats';
 import { Client } from './client';
+import { Blockchain } from '../types/blockchain';
 
 export class Stats extends GenericStats<Client> {
     public getAccountDelegateStats(): AccountStats {
```
krisboit, line 12: this needs to be async, when we will implement this we will need to do some http requests

src/screens/token/components/delegate-token/components/tabs/account-tab/account-tab.tsx
```diff
@@ -0,0 +1,99 @@
...
+};
+
+export class AccountTabComponent extends React.Component<
+    IProps & IReduxProps & IThemeProps<ReturnType<typeof stylesProvider>>
+> {
+    public render() {
+        const styles = this.props.styles;
+
+        const blockchainInstance = getBlockchain(this.props.blockchain);
+        const stats = blockchainInstance.getStats(this.props.chainId);
```
krisboit, line 45: try no to do heavy computation in render functions, render function can be called many times. In this case stats will  have only async methods, so we need to use hooks, or transform this in class and use lifecycle methods.

src/screens/token/components/delegate-token/components/tabs/account-tab/styles.ts
```diff
@@ -0,0 +1,35 @@
...
+            paddingHorizontal: BASE_DIMENSION,
+            marginTop: BASE_DIMENSION * 3,
+            paddingBottom: BASE_DIMENSION * 5
+        },
+        text: {
+            fontWeight: 'bold',
+            fontSize: normalize(22),
+            lineHeight: normalize(28),
+            textAlign: 'center',
+            letterSpacing: 0.35,
```
krisboit, line 19: i think we should use normalise here. @NovacAndrei any thoughts ?

src/screens/token/components/delegate-token/components/tabs/validators-tab/styles.ts
```diff
@@ -18,5 +23,9 @@ export default (theme: ITheme) =>
             letterSpacing: 0.35,
```
krisboit, line 23: same here, shouldn't we use normalise here ?

src/screens/token/components/delegate-token/components/validators/validator-card/validator-card.tsx
```diff
@@ -0,0 +1,102 @@
...
+    bottomStats: IStatValue[];
+    blockchain: Blockchain;
+}
+
+export function getValueString(stat: IStatValue) {
+    switch (stat.type) {
+        case IStatValueType.STRING:
+            return stat.data.value;
+        case IStatValueType.AMOUNT:
+            return stat.data.value + ' ' + stat.data.tokenSymbol; // TODO format text based on blockchain
```
krisboit, line 33: maybe the value should also be formatted here, value should be in atomic units

src/screens/token/components/delegate-token/components/validators/validator-card/validator-card.tsx
```diff
@@ -0,0 +1,102 @@
...
+export const ValidatorCardComponent = (
+    props: IExternalProps & IThemeProps<ReturnType<typeof stylesProvider>>
+) => {
+    return (
+        <TouchableHighlight
+            onPress={() => {
+                NavigationService.navigate('Validator', {
+                    blockchain: props.blockchain,
+                    validator:
+                        props.blockchain === Blockchain.CELO
```
krisboit, line 46: is this necessary ? In all these components conditionals based on blockchain should be avoided

src/core/blockchain/celo/config.ts
NovacAndrei: Understood, well it makes sense.

src/screens/token/components/delegate-token/components/tabs/account-tab/styles.ts
NovacAndrei: Hmm true. Well we use in the entire app almost the same letterSpacing value. We should move this in the same place as BASE_DIMENSION and avoid hardcoding it any more. 

But I think I will refactor this in a separate PR because I would like to change in the entire app everything with letterSpacing.

src/screens/token/components/delegate-token/components/validators/validator-card/validator-card.tsx
georgeburduhos: It was just for dummy data not to confuse the Cosmos labels. Moved it in stats where there are multiples stats dummy

PR #728, opened 2020-05-14
Extension Fix Preferences Blockchains

NovacAndrei:
- fix Blockchain portfolio order and active on web
- also fix a screen height on Dashboard web when bottom blockchain nav is hidden


src/core/connect-extension/conn-ext-trim-state.ts
```diff
@@ -106,7 +106,7 @@ const trimPreferences = (preferences: IPrefState): IExtStorage.IStoragePreferenc
         currency: preferences.currency,
         testNet: preferences.testNet,
         networks: cloneDeep(preferences.networks),
-        blockchains: Object.keys(preferences.blockchains)
+        blockchains: cloneDeep(preferences.blockchains)
```
krisboit, line 109: Since we change the type here, shouldn't we do changes on extension side, where we load this data in state.
Also is this change backward compatible ?

src/core/connect-extension/types.ts
```diff
@@ -107,7 +107,12 @@ export interface IStoragePreferences {
     currency: string;
     testNet: boolean;
     networks: IStorageNetworks;
-    blockchains: string[];
+    blockchains: {
```
krisboit, line 110: this can remain as it was, we just need to filter the list and order it

PR #734, opened 2020-05-15
Handle Sentry Errors

NovacAndrei:


src/core/secure/keychain.ts
```diff
@@ -90,7 +91,8 @@ export const verifyPinCode = async (pinCode: string): Promise<boolean> => {
         const encryptionKey = await getEncryptionKey(pinCode);
         await readEncrypted(KEY_PIN_SAMPLE, encryptionKey);
         return true;
-    } catch (e) {
+    } catch (err) {
+        Sentry.captureException(new Error(JSON.stringify(err))); // needed?
```
krisboit, line 95: this would be ok, but if we can filter keystore errors, we don't want to know that a user mistyped his pin code. So i think we can remove this.

src/app.tsx
```diff
@@ -183,6 +184,14 @@ export default class App extends React.Component<{}, IState> {
                                 theme="dark"
                                 onNavigationStateChange={(_, newState) => {
                                     this.setState({ navigationState: newState });
+
+                                    const currentRoute = NavigationService.getCurrentRouteWithParams();
```
krisboit, line 188: we could keep current route object at app level, and if it's changed we trigger addBreadcrumb, if not, we do nothing

PR #744, opened 2020-05-18
Extension Dahsboard menu remove QR code

NovacAndrei:


NovacAndrei:
@georgeburduhos Please don't review this yet, it seems that there are some problems with the ext

PR #750, opened 2020-05-20
Fix sentry errors

krisboit:


src/screens/wallets/wallets.tsx
```diff
@@ -240,6 +240,8 @@ export class WalletsScreenComponent extends React.Component<
         const styles = this.props.styles;
         const { selectedTab } = this.state;
 
+        // console.log(this.props.wallets[WalletType.HD].map(w => w.name));
```
NovacAndrei, line 243: I think this comment is not needed

PR #751, opened 2020-05-20
QR Code Transactions support - Gitcoin

NovacAndrei:
Zilliqa ZIP: https://github.com/Zilliqa/ZIP/blob/master/zips/zip-7.md

Pattern to parse zilliqa url based on ZIP-7: https://regexr.com/54nof

Generating QR code: https://www.the-qrcode-generator.com/

XSGD contract address: `zil1n0006zrsdtl0zj5mwac2rkaa442f4d37hntkv7`


src/screens/transaction-request/qr-code-tx-request/qr-code-tx-request.tsx
```diff
@@ -0,0 +1,627 @@
...
+    }
+
+    public async parseQrCodeTxPayload() {
+        const { blockchain } = this.props.selectedAccount;
+        const { qrCodeTxPayload } = this.props;
+        const blockchainInstance = getBlockchain(blockchain);
+
+        this.setState({ isLoading: true });
+
+        if (qrCodeTxPayload?.chainId) {
```
krisboit, line 180: this is not quite ok...
chainId could be mainnet chainID, and here you are assuming that chainID could be only testnet

src/screens/transaction-request/qr-code-tx-request/qr-code-tx-request.tsx
```diff
@@ -0,0 +1,627 @@
...
+            if (this.isChainIdValid(qrCodeTxPayload.chainId)) {
+                // Valid ChainId
+                // Set the ChainId of the user with the received one
+                this.props.setNetworkTestNetChainId(blockchain, Number(qrCodeTxPayload.chainId));
```
krisboit, line 184: chainId should not be converted into number, we should keep things generic

src/screens/transaction-request/qr-code-tx-request/qr-code-tx-request.tsx
```diff
@@ -0,0 +1,627 @@
...
+            this.setState({ tokenSymbol: tokenSymbolOnChainId });
+            this.getAccountTokenBySymbol(tokenSymbolOnChainId);
+        }
+    }
+
+    private async fetchTokenByContractAddress(address: string): Promise<string> {
+        let foundTokenSymbol: string;
+
+        try {
+            const token = await fetch(
```
krisboit, line 297: I think it would be better to search for the token in blockchain.

src/screens/transaction-request/qr-code-tx-request/qr-code-tx-request.tsx
```diff
@@ -178,18 +179,26 @@ export class QRCodeTransferRequestComponent extends React.Component<
...
                 // Valid ChainId
                 // Set the ChainId of the user with the received one
-                this.props.setNetworkTestNetChainId(blockchain, Number(qrCodeTxPayload.chainId));
+                this.props.setNetworkTestNetChainId(blockchain, chainId);
 
-                // Activate testnet if disabled
-                if (this.props.isTestNet === false) {
-                    // Activate TestNet
-                    this.props.toggleTestNet();
+                if (this.isChainIdMainNet(chainId)) {
```
krisboit, line 189: You could simplify this condition:
```ts
if (!this.isChainIdMainNet(chainId) !== this.props.isTestNet) {
    this.props.toggleTestNet();
}
```

PR #752, opened 2020-05-20
More extension enhancements

krisboit:


src/core/connect-extension/conn-ext-build-state/build-preferences.ts
```diff
@@ -14,6 +15,14 @@ export const buildPreferences = (trimmedPreferences: IExtStorage.IStoragePrefere
         });
     });
 
+    // add inactive blockchains
+    BLOCKCHAIN_LIST.filter(b => !trimmedPreferences.blockchains.includes(b)).map(blockchain => {
+        blockchains[blockchain] = {
+            order: 999,
```
NovacAndrei, line 21: Should we use 
```
 BLOCKCHAIN_LIST.filter(b => !trimmedPreferences.blockchains.includes(b)).map((blockchain, index) => {
        blockchains[blockchain] = {
            order: 999 - index,
            active: false
        };
    });
```
So that each blockchain will have an unique order? 
Or it's not necessary.

src/redux/preferences/selectors.ts
```diff
@@ -1,4 +1,4 @@
-import { Platform } from 'react-native';
+// import { Platform } from 'react-native';
```
NovacAndrei, line 1: Import, I think it's not needed

src/core/connect-extension/conn-ext-build-state/build-preferences.ts
krisboit: it's not necessary, actually the order could be random, since everything in that list will be inactive :)

Issue #733, opened 2020-05-15
Wallet is not switched if i make a transaction on extension


georgeburduhos:
Steps:

1. Create 2 wallets - in my case i had HD wallet and ledger
2. Set active wallet on both to ledger
3. Switch from extension to hd wallet
4. Initiate a transaction
5. Confirm signing after push is received on mobile.

Result - it asks to confirm the transaction on ledger but is should sign a transaction since it comes from hd wallet

krisboit:
@georgeburduhos did you have the same mnemonic on both HD wallet and ledger ?

krisboit:
i may found the problem, on send screen we should send the walletID (https://github.com/Moonlet/wallet-app/blob/master/src/screens/send/send.tsx#L154). Right now walletName is sent, but it's just for display purpose, we should send wallet id and use it to sign from the correct wallet.

georgeburduhos:
@krisboit  yes it was the same mnemonic on both HD wallet and ledger

PR #756, opened 2020-05-25
Web Ext Transaction Request (Switch wallet) & Refactor Letter Spacing

NovacAndrei:
- Wallet is not switched on ext Tx Request #733
- Refactor Letter Spacing #732


src/screens/transaction-request/transaction-request.tsx
```diff
@@ -79,7 +87,21 @@ export class TransactionRequestScreenComponent extends React.Component<
...
 
             if (payload) {
-                this.setState({ moonletTransferPayload: payload });
+                const walletId = payload.walletId;
+                if (walletId !== this.props.selectedWallet.id) {
+                    if (
+                        Object.keys(this.props.wallets).filter(wId => wId === walletId).length === 1
+                    ) {
+                        // Switch the wallet
+                        this.props.setSelectedWallet(walletId);
```
georgeburduhos, line 96: What happens if on the new selected wallet the selected blockchain is different?

NovacAndrei: It's safe because the `blockchain` is taken from the `account` which is send via transfer payload.
It's not based on the phone's selected blockchain.

PR #763, opened 2020-05-26
Preparation for multiple web platforms

krisboit:


src/core/connect-extension/connect-extension-web.extension.ts
```diff
@@ -1,9 +1,9 @@
-import { generateRandomEncryptionKey, decrypt } from '../secure/encrypt/encrypt.web';
+import { generateRandomEncryptionKey, decrypt } from '../secure/encrypt/encrypt.extension';
 import { v4 as uuidv4 } from 'uuid';
 import { IQRCodeConn, FirebaseRef, IStorage } from './types';
 import { storage, database } from 'firebase';
-import { storeEncrypted, readEncrypted, deleteFromStorage } from '../secure/storage/storage.web';
-import { CONN_EXTENSION } from '../../core/constants/app';
+import { storeEncrypted, readEncrypted, deleteFromStorage } from '../secure/storage/storage';
```
NovacAndrei, line 5: We are going to have a problem if we do this import, because we encrypt from mobile and decrypt on web. For now these functions have different implementation.

For now we should import:
```
import { storeEncrypted, readEncrypted, deleteFromStorage } from '../secure/storage/storage.extension';
```
Until we are going to implement the same encryption on both mobile and extension and do the encrypt migrations.

NovacAndrei: I have updated this 🙂

PR #791, opened 2020-06-02
Setup E2E tests with Detox - Beta

NovacAndrei:


src/library/button/button.tsx
```diff
@@ -27,11 +27,17 @@ export interface IButtonProps {
     onPressIn?: any;
     onPressOut?: any;
     bottomLabel?: string;
+    testIDWrapper?: string;
```
georgeburduhos, line 30: I think we should enforce this. It will force us to set testId for all future screens 
* in a future separate PR :)

PR #797, opened 2020-06-09
locked gold contract implementation

georgeburduhos:


src/redux/wallets/actions/pos-actions.ts
```diff
@@ -0,0 +1,139 @@
...
+import { LedgerWallet } from '../../../core/wallet/hw-wallet/ledger/ledger-wallet';
+import { TRANSACTION_PUBLISHED } from '../actions';
+import { ConnectExtension } from '../../../core/connect-extension/connect-extension';
+import { CLOSE_TX_REQUEST, closeTransactionRequest } from '../../ui/transaction-request/actions';
+import { translate } from '../../../core/i18n';
+import { Dialog } from '../../../components/dialog/dialog';
+import { PosBasicActionType } from '../../../core/blockchain/types/token';
+import { IValidator } from '../../../core/blockchain/types/stats';
+
+export const quickDelegate = (
```
krisboit, line 28: by looking at this, i think we could do it in an abstract way, the only difference between all transaction is at line `64 const txs = await blockchainInstance.transaction.buildPosTransaction(...`
For now you can leave it like this, but ew have to talk about it, i'm sure we can find a solution :)

georgeburduhos: Yes. only one thing will be different in most of this actions

PR #798, opened 2020-06-09
Connect/Pair Ext with Phone - Retry and Sentry Errors

NovacAndrei:


src/core/connect-extension/connect-extension-web.extension.ts
```diff
@@ -195,49 +197,92 @@ export const ConnectExtensionWeb = (() => {
...
 
-                        // Store connection
-                        await storeConnection(conn);
+            // close loading modal
+            await LoadingModal.close();
 
-                        // remove listener for connectionId
-                        connectionsRef.child(conn.connectionId).off('value');
+            // Reset sync conn attempts
+            syncConnAttempts = 0;
```
krisboit, line 235: to be sure that flow will be fine you should initialise `syncConnAttempts` with `0` at the beginning of the flow, it would be more simple, but by using a global variable here if there are calls to `syncConnect` from multiple places your code will not work correctly.

To be more safe you should avoid using global vars (`syncConnAttempts`), you can achieve this by using a recursive function. e.g.
```ts
const syncConnect = async (conn: IQRCodeConn, connectionsRef: any, syncConnAttempts: number = CONN_EXT_RETRY_ATTEMPTS) => {
  try {
    if (syncConnAttempts > 0) {
      //do stuff
    } else {
      // handle error
    }
  } catch (e) {
    // something happened
    // check if retry is needed
    syncConnect(conn, connectionsRef, syncConnAttempts - 1);
  }
}

PR #809, opened 2020-06-18
Fix Security Checks - continue label

NovacAndrei:
- sentry error


src/components/security-checks/security-checks.tsx
```diff
@@ -101,35 +101,36 @@ export class SecurityChecksComponent extends React.Component<
...
-                                {translate('App.labels.continue')}
-                            </Button>
-                        </View>
+                        <Button
+                            testID="button-understand"
+                            wrapperStyle={styles.bottomButton}
+                            primary
+                            onPress={() => this.continue()}
+                        >
+                            {'Continue'}
```
krisboit, line 132: maybe if we use Translate component, the problem would be fixed, or just wrap the text in Text component

PR #822, opened 2020-06-30
Notification Center - API Calls

NovacAndrei:
TODO:
- after API deploy, update `walletApiBaseUrl`
- refactor `NotificationType`


src/config/config-beta.ts
```diff
@@ -22,5 +24,13 @@ export const CONFIG_BETA: IAppConfig = {
         sendResponseUrl: extSyncBaseUrl + '/sendResponse',
         deleteRequestUrl: extSyncBaseUrl + '/deleteRequest'
     },
-    firebaseWebConfig: firebaseConfig
+    firebaseWebConfig: firebaseConfig,
+    notificationCenter: {
```
krisboit, line 28: i know we already used this pattern, but it's not really scalable, since we need to add all our endpoints in this config. So, a better approach would be to have only the `walletApiBaseUrl` in this config, the rest could stay in a specific client

src/screens/dashboard/dashboard.tsx
```diff
@@ -209,6 +239,77 @@ export class DashboardScreenComponent extends React.Component<
...
+            this.props.walletId !== prevProps.walletId &&
+            isFeatureActive(RemoteFeature.NOTIF_CENTER)
+        ) {
+            this.hasUnseenNotifications();
+        }
+    }
+
+    private async hasUnseenNotifications() {
+        try {
+            const http = new HttpClient(CONFIG.notificationCenter.hasUnseenNotifs);
```
krisboit, line 253: since we use redux, we should do all http calls in actions, this way the logic of the screen remains simple, and also actions are reusable...

src/screens/notifications/notifications.tsx
```diff
@@ -70,8 +137,78 @@ export class NotificationsComponent extends React.Component<
...
+            //
+        }
+    }
+
+    private async fetchNotifications() {
+        const { page, notifications } = this.state;
+
+        try {
+            // Fetch next page
+            const http = new HttpClient(
```
krisboit, line 164: same here, let's use actions for this, also by using redux for this we cache the notifications, we minimise the calls to our api. The redux property should not be persisted.

src/core/messaging/types.ts
```diff
@@ -24,3 +26,8 @@ export interface ISilentMessage {
     type: SilentMessageType;
     data: any;
 }
+
+export enum PushNotifTokenType {
+    FCM = 'fcm',
```
krisboit, line 31: FCM = 'FCM',

src/core/messaging/types.ts
```diff
@@ -1,9 +1,11 @@
+// TODO: refactor this, too many types
 export enum NotificationType {
     EXTENSION_TRANSACTION = 'EXTENSION_TRANSACTION',
```
krisboit, line 3: We need to cleanup this enum ,,, what's the difference between EXTENSION_TRANSACTION and EXTENSION_SIGN_TX or MOONLET_TRANSFER

src/redux/notifications/actions.ts
```diff
@@ -1,3 +1,180 @@
...
+import { getTokenConfig } from '../tokens/static-selectors';
+
+export const SET_HAS_UNSEEN_NOTIFICATIONS = 'SET_HAS_UNSEEN_NOTIFICATIONS';
+export const SET_NOTIFICATIONS = 'SET_NOTIFICATIONS';
+export const MARK_SEEN = 'MARK_SEEN';
+
+const setHasUnseenNotifications = (hasUnseenNotifications: boolean) => {
+    return {
+        type: SET_HAS_UNSEEN_NOTIFICATIONS,
+        data: { hasUnseenNotifications }
```
krisboit, line 20: instead of having hasUnseenNotifications as a boolean, maybe we should have unseenNotifications as a number. it may be easier on the UI

src/redux/notifications/actions.ts
```diff
@@ -1,3 +1,267 @@
...
+};
+
+export const setNotifications = (notifications: any) => {
+    return {
+        type: SET_NOTIFICATIONS,
+        data: { notifications }
+    };
+};
+
+const getTimestamp = () => {
```
georgeburduhos, line 33: ntp server i think would be a better idea

src/redux/notifications/actions.ts
```diff
@@ -1,3 +1,267 @@
...
+    return moment().unix() * 1000;
+};
+
+const getDomain = () => {
```
georgeburduhos, line 37: we might use this for other functions in api, make an utils where the api is implemented

src/redux/notifications/actions.ts
```diff
@@ -1,3 +1,267 @@
...
+            const signature = getSignature(data, walletPrivateKey, walletPublicKey);
+            data.walletPublicKeys = [
+                {
+                    walletPublicKey,
+                    signature
+                }
+            ];
+
+            const http = new HttpClient(CONFIG.walletApiBaseUrl);
+            const res = await http.post('/notifications/unseen', data);
```
georgeburduhos, line 76: I think a better idea is to make a Moonlet-api client for this and just implement functions for each endpoint

src/core/messaging/handlers/notification.ts
```diff
@@ -69,7 +80,8 @@ const handleNotification = (
             if (requestId) {
                 store.dispatch(openTransactionRequest({ requestId }));
             } else {
-                //
+                // maybe find a way to handle this
```
georgeburduhos, line 83: at least send something in sentry for us to be aware that this happens sometimes

index.js
```diff
@@ -40,6 +40,11 @@ import androidBgMessagingHandler from './src/core/messaging/silent/android-backg
 import { YellowBox } from 'react-native';
 YellowBox.ignoreWarnings(['RCTRootView cancelTouches', 'Require cycle']);
 
+// see https://github.com/facebook/react-native/issues/16434
+import { URL, URLSearchParams } from 'whatwg-url';
```
georgeburduhos, line 44: What impact/problem does this have?

src/redux/notifications/reducer.ts
```diff
@@ -1,13 +1,65 @@
...
+                    body: notif.body,
+                    seen: notif.seen,
+                    data: notif.data
+                };
+
+                notifications = {
+                    ...notifications,
+                    [notif.data.blockchain]: {
+                        ...notifications[notif.data.blockchain],
+                        [notif._id]: notifData
```
georgeburduhos, line 37: is ok to use notif._id ? - _id is not something only used by mongo, and we use notifId?

src/redux/wallets/actions/wallet-actions.ts
```diff
@@ -52,14 +52,17 @@ import {
...
+    getWalletCredentialsKey,
+    setWalletCredentialsKey
 } from '../../../core/secure/keychain/keychain';
 import { delay } from '../../../core/utils/time';
 import { toggleBiometricAuth } from '../../preferences/actions';
 import { CLOSE_TX_REQUEST, closeTransactionRequest } from '../../ui/transaction-request/actions';
 import { ConnectExtension } from '../../../core/connect-extension/connect-extension';
 import { LoadingModal } from '../../../components/loading-modal/loading-modal';
 import { captureException as SentryCaptureException } from '@sentry/react-native';
+import { store } from '../../config';
```
georgeburduhos, line 65: Remove this :)

src/redux/wallets/actions/wallet-actions.ts
```diff
@@ -778,3 +787,45 @@ export const addPublishedTxToAccount = (
...
+
+    for (const wallet of Object.values(state.wallets)) {
+        try {
+            const walletId = (wallet as IWalletState).id;
+
+            const storageWallet = await HDWallet.loadFromStorage(walletId, password);
+            const walletCredentials = storageWallet.getWalletCredentials();
+
+            if (walletCredentials) {
+                store.dispatch(setWalletPublicKey(walletId, walletCredentials.publicKey));
```
georgeburduhos, line 812: you can use 

` setWalletPublicKey(walletId, walletCredentials.publicKey)(dispatch, getState);`

src/redux/wallets/state.ts
```diff
@@ -8,6 +8,7 @@ export interface IWalletsState {
 
 export interface IWalletState {
     id: string;
+    walletPublicKey?: string;
```
georgeburduhos, line 11: shouldn't we enforce this?

src/screens/dashboard/dashboard.tsx
```diff
@@ -63,12 +71,19 @@ export interface IReduxProps {
     selectedBlockchainAccounts: IAccountState[];
     userCurrency: string;
     chainId: ChainIdType;
+    setNotifications: typeof setNotifications;
+    deviceId: string;
+    fetchNotifications: (page?: number) => Promise<any>;
```
georgeburduhos, line 76: Too many actions specific for notifications on this screen that are not actually used

create just one action that handles notifications (register, fetch, getUnseen) and here just show the info from redux (red bullet on the notifications icon)

src/screens/notifications/notifications.tsx
```diff
@@ -85,12 +242,42 @@ export class NotificationsComponent extends React.Component<
...
+                        currentOffset >= this.scrollOffset
+                            ? this.setState({ hideBottomNav: true }) // scroll down
+                            : this.setState({ hideBottomNav: false }); // scroll up
+                        this.scrollOffset = currentOffset;
+
+                        if (currentOffset === 0) {
+                            this.setState({ hideBottomNav: false }); // scroll on top
+                        }
+                    }}
+                    refreshControl={
```
georgeburduhos, line 265: We should find a way to show user that has this option - like in manage wallets

index.js
NovacAndrei: default `URL` is not implemented on iOS, throws an error, so we need to change the base import of this
I think `URL` and `URLSearchParams` are quite useful to use, they'll help us a lot when parsing urls

@krisboit what do you think about this?

NovacAndrei: I have switched to `string.replace`

PR #825, opened 2020-07-03
Fix extension build tokens state

NovacAndrei:


src/core/connect-extension/conn-ext-build-state/build-tokens.ts
```diff
@@ -39,21 +36,27 @@ const fetchToken = async (
...
                 .getClient(chainId)
-                .tokens[tk.type].getTokenInfo(tk.contractAddress);
+                .tokens[tk.type].getTokenInfo(tk.contractAddress)
+        ]);
+        if (getTokenInfo[0].status === 200) {
+            staticToken = await getTokenInfo[0].json();
         }
+        blockchainToken = getTokenInfo[1];
     } catch {
         //
```
krisboit, line 53: From now on no more empty catches :) Let's take an action if something fails.

PR #835, opened 2020-07-15
Web wallet frame1

standreinmcp:


package.json
```diff
@@ -105,6 +108,7 @@
         "react-native-qrcode-svg": "^6.0.3",
         "react-native-randombytes": "^3.5.3",
         "react-native-reanimated": "^1.7.0",
+        "react-native-safe-area-context": "^3.1.1",
```
NovacAndrei, line 111: Is this used?

src/components/web-components/rank-card/styles.ts
```diff
@@ -0,0 +1,37 @@
+import { StyleSheet } from 'react-native';
+import { ITheme } from '../../../core/theme/itheme';
+
+export default (theme: ITheme) =>
+    StyleSheet.create({
+        container: {
+            backgroundColor: theme.colors.rankCard,
+            marginRight: '16.84%',
```
georgeburduhos, line 8: Use some consts here as much as possible - like BASE_DIMENSION in wallet app and adjust the values to something similar: like make a BASE_DIMENSION = 4. and for something like this you can use BASE_DIMENSION *4

src/components/web-components/rank-card/styles.ts
```diff
@@ -0,0 +1,37 @@
...
+            marginLeft: '16.84%',
+            marginTop: 37,
```
georgeburduhos, line 10: same here something like BASE_DIMENSION * 8 
or you can make BASE_DIMENSION higher since the values are pretty big

src/components/web-components/rank-card/rank-card.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { View, Text } from 'react-native';
+import stylesProvider from './styles';
+import { IThemeProps, withTheme } from '../../../core/theme/with-theme';
+
+export const RankCardComponent = (props: IThemeProps<ReturnType<typeof stylesProvider>>) => {
```
georgeburduhos, line 6: rename this to TopStatsCard

src/screens/connect-web/styles.ts
```diff
@@ -0,0 +1,68 @@
+import { StyleSheet } from 'react-native';
+import {
+    // SCREEN_HEIGHT,
```
georgeburduhos, line 3: delete if unused

src/components/web-components/rank-card/rank-card.tsx
NovacAndrei, line 6: I thins component, you should send the hardcoded values as ExternalProps.

Also you can make this component a little smarter and so that you don't duplicate a lot of code.
This is only an example, you choose what labels and values helps you. 

```
export interface IExternalProps {
    title1: string;
    value1: string;

    title2: string;
    value2: string;

    ...
}

export const RankCardComponent = (
    props: IExternalProps & IThemeProps<ReturnType<typeof stylesProvider>>
) => {
    const { styles } = props;

    const renderRow = (title: string, subtitle: string) => {
        return (
            <View style={styles.column}>
                <Text style={styles.title}>{title}</Text>
                <Text style={styles.value}>{subtitle}</Text>
            </View>
        );
    };

    return (
        <View style={styles.container}>
            {renderRow(props.title1, props.value1)}
            {renderRow(props.title2, props.value2)}

            ...
        </View>
    );
};

export const RankCard = smartConnect<IExternalProps>(RankCardComponent, [
    withTheme(stylesProvider)
]);
```

src/components/web-components/rank-card/styles.ts
NovacAndrei, line 10: It's a best practice not to use hardcoded values. 
We should build a function like `normalize` and then use `BASE_DIMENSION` based on that.

Now, on mobile we have the scale based on iPhone 11, but we should find a way to also normalize on web. We need to make a little research about best practices of normalize on web.

```
// based on iPhone 11's scale
const scale = SCREEN_WIDTH / 414;
```

src/components/web-components/rank-card/styles.ts
standreinmcp: Okay. But when i do that, it's not responding to resizing the browser window

src/components/web-components/rank-card/styles.ts
NovacAndrei: Also, as a rule of thumb in design, please build your items from top to to bottom, meaning that you should only use `marginBottom`. Only if necessary, add `marginTop`

src/components/web-components/rank-card/styles.ts
georgeburduhos: Then maybe make a const as percentage. The idea is that this value will be almost everywhere the same margin

src/components/web-components/rank-card/styles.ts
```diff
@@ -0,0 +1,37 @@
...
+            height: 67,
```
NovacAndrei, line 11: Also this is not quite scalable, you should use `paddingVertical` instead of `height`.

Also it would be easier if we use multiple of 8 as measurement.

src/components/web-components/rank-card/styles.ts
```diff
@@ -0,0 +1,37 @@
...
+            borderRadius: 6
+        },
+        column: {
+            flex: 1,
+            justifyContent: 'center',
+            alignItems: 'center'
+        },
+        title: {
+            fontSize: 16,
+            fontWeight: 'normal',
```
NovacAndrei, line 22: Not needed, normal is `default`

src/components/web-components/top-bar/styles.ts
```diff
@@ -0,0 +1,63 @@
...
+import { ITheme } from '../../../core/theme/itheme';
+
+export default (theme: ITheme) =>
+    StyleSheet.create({
+        container: {
+            flex: 1
+        },
+        headerContainer: {
+            flexDirection: 'row',
+            height: 50
```
NovacAndrei, line 12: Same here, please use paddingVertical and then same normalized values.

src/components/web-components/top-bar/styles.ts
```diff
@@ -0,0 +1,63 @@
...
+            paddingLeft: 20
+        },
+        logoText: {
+            color: theme.colors.accent
+        },
+        headerCenterText: {
+            color: theme.colors.text,
+            fontStyle: 'normal',
+            fontWeight: 'bold',
+            fontSize: normalizeFontAndLineHeight(22),
```
NovacAndrei, line 51: We should not use this on web, because `normalizeFontAndLineHeight` uses react native components.
We should build something specific for web.

src/components/web-components/top-bar/styles.ts
```diff
@@ -0,0 +1,63 @@
...
+            lineHeight: normalizeFontAndLineHeight(41),
+            letterSpacing: 0.4
+        },
+        socialContainer: {
+            borderLeftColor: theme.colors.separator,
+            borderLeftWidth: 2,
+            display: 'flex',
+            flexDirection: 'row',
+            paddingTop: 12,
```
NovacAndrei, line 60: Same here, build from top to bottom.

src/components/web-components/top-bar/styles.ts
```diff
@@ -0,0 +1,63 @@
...
+            height: '100%'
```
NovacAndrei, line 61: You also have `flex : 1` which helps you, instead of height and width. Check here if helps.

src/screens/connect-web/connect-web-wallet.tsx
```diff
@@ -0,0 +1,58 @@
...
+                <Text style={styles.subTitle}>${state.amount2}</Text>
+            </View>
+
+            <RankCard />
+
+            <View style={styles.connectLedgerContainer}>
+                <TouchableOpacity style={{ width: '100%' }}>
+                    <View style={styles.connectLedgerButton}>
+                        <Icon name={IconValues.LEDGER_LOOGO} size={normalize(6)} />
+                        <Text style={styles.connectLedgerText}>Connect Ledger</Text>
```
NovacAndrei, line 50: Also I think we should implement and use `translate` on web. 
Maybe in near future we are going to use multiple languages.

src/screens/connect-web/connect-web-wallet.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { IconValues } from '../../components/icon/values';
+import { normalize } from '../../styles/dimensions';
+import { withTheme, IThemeProps } from '../../core/theme/with-theme';
+import stylesProvider from './styles';
+
+export const ConnectWebWalletScreenComponent = (
+    props: IThemeProps<ReturnType<typeof stylesProvider>>
+) => {
+    const styles = props.styles;
+    const [state, setState] = useState({
```
NovacAndrei, line 16: If you want to have more control of the types, you can use something like:

```
const [textTop, setTextTop] = useState<string>('');
```

PR #855, opened 2020-07-23
Ledger Wallets Credentials

NovacAndrei:


src/redux/wallets/actions/wallet-actions.ts
```diff
@@ -819,13 +843,24 @@ export const setWalletsCredentials = (password: string) => async (
                 );
 
                 if (keychainWalletCredentials) {
-                    // already set
+                    // Already set
```
krisboit, line 846: you can get rid of these empty block by using negations. e.g.:
```ts
if (!keychainWalletCredentials) {
...
}
```

PR #872, opened 2020-08-03
Updates Notification Center

NovacAndrei:


src/redux/wallets/actions/wallet-actions.ts
```diff
@@ -866,3 +866,21 @@ export const setWalletsCredentials = (password: string) => async (
...
+    for (const wallet of Object.values(state.wallets)) {
+        for (const account of wallet.accounts) {
+            if (account.address?.toLocaleLowerCase() === address?.toLocaleLowerCase()) {
+                return {
+                    walletName: wallet.name,
+                    accountName: account.name || `Account ${account.index + 1}`
+                };
+            }
+        }
+    }
```
krisboit, line 885: you should return something if nothing is found.

src/core/utils/api-client/notifications.ts
```diff
@@ -22,25 +23,37 @@ export class NotificationsApiClient {
...
+
+                const walletData: any = {
                     walletPublicKey,
                     timestamp: await getCurrentTimestampNTP(),
-                    domain: getWalletApiDomain(),
-                    page: page || 1
+                    domain: getWalletApiDomain()
                 };
 
                 const signature = getSignature(data, walletPrivateKey, walletPublicKey);
```
krisboit, line 47: you should use walletData for sign, right ? it's weird to sign all data object...

src/core/utils/api-client/notifications.ts
```diff
@@ -22,25 +23,37 @@ export class NotificationsApiClient {
...
+                blockchainNetworks,
+                page: page || 1
+            };
+
+            for (const walletPublicKey of walletPublicKeys) {
+                const walletPrivateKey = await getWalletCredentialsKey(walletPublicKey);
+
+                const walletData: any = {
                     walletPublicKey,
                     timestamp: await getCurrentTimestampNTP(),
```
krisboit, line 43: does this have a cache ? or it does a ntp call every time ?

src/core/utils/api-client/notifications.ts
```diff
@@ -175,24 +181,28 @@ export class NotificationsApiClient {
...
+                const walletData: any = {
                     walletPublicKey,
-                    signature
-                }
-            ];
+                    timestamp: await getCurrentTimestampNTP(),
+                    domain: getWalletApiDomain()
+                };
+
+                const signature = getSignature(data, walletPrivateKey, walletPublicKey);
```
krisboit, line 201: same here, only walletData should be signed

src/core/utils/api-client/notifications.ts
NovacAndrei: It does make a ntp call every time.
Any suggestions on how to cache?

krisboit: yep, quite simple: `const timestamp = await getCurrentTimestampNTP()` before `for` :) this way you do it once :P

PR #875, opened 2020-08-05
Fix Open Notification ChainId Type

NovacAndrei:


src/core/messaging/handlers/notification.ts
```diff
@@ -45,6 +46,22 @@ const handleNotification = (
         case NotificationType.TRANSACTION:
             const data = JSON.parse(notification.data);
 
+            let chainId = data.chainId;
+
+            // TODO: remove this on chainId refactor
+            switch (data.blockchain) {
```
krisboit, line 52: we have to avoid this kind of `switch`. I found the real problem :), we had it on Zilliqa too. The fix has to be done in `src/core/blockchain/ethereum/index.ts` in `getClient` function (check the solution on Zilliqa).

src/screens/notifications/notifications.tsx
```diff
@@ -128,9 +128,11 @@ export class NotificationsComponent extends React.Component<
                 break;
         }
 
-        const { notifications } = this.state;
-        notifications[notificationId].seen = true;
-        this.setState({ notifications });
+        const notifIndex = notifications.findIndex(notif => notif.id === notificationId);
```
krisboit, line 131: you don't really need the index, use `find` instead (check this out: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find)

krisboit: everything will be more simple :)

src/screens/notifications/notifications.tsx
```diff
@@ -90,20 +90,20 @@ export class NotificationsComponent extends React.Component<
     public async componentDidMount() {
         await LoadingModal.open();
```
krisboit, line 91: here you should call `fetchNotifications` and that's it. Everything else should be done there, even the modal open/close.

src/core/messaging/handlers/notification.ts
NovacAndrei: Yes, nice, I had remember and updated :)

src/screens/notifications/notifications.tsx
NovacAndrei: What do you think about this:

```
this.setState({
            notifications: this.state.notifications.map(notif =>
                notif.id === notificationId
                    ? {
                          ...notif,
                          seen: true
                      }
                    : notif
            )
        });

```

I think this is easier :)

PR #884, opened 2020-08-14
Extension sdk

krisboit:


src/app.tsx
```diff
@@ -180,42 +181,46 @@ export default class App extends React.Component<{}, IState> {
 
     public render() {
         if (this.state.appReady) {
+            const bgRequest = !!document.location.hash;
```
NovacAndrei, line 184: I think we should check this for mobile, it returns undefined for `document.location.hash`. 
Also because of this the tests are crashing and the mobile app is crashing, cannot find `document`.

NovacAndrei: If all good, I have updated it to:

```
const bgRequest = Platform.OS === 'web' && !!document.location.hash;
```

PR #926, opened 2020-09-16
Implemented Ethereum name service

standreinmcp:


src/core/blockchain/types/name-service.ts
```diff
@@ -24,5 +24,5 @@ export interface IResolveTextResponse {
 
 export interface IBlockchainNameService {
     resolveText(text: string): Promise<IResolveTextResponse>;
-    resolveName(text: string): Promise<IResolveNameResponse>;
+    resolveName(text: string): Promise<string | IResolveNameResponse>;
```
georgeburduhos, line 27: Always respond with IResolveNameResponse and add address inside the object

krisboit (changes requested):

src/screens/send/send.tsx
```diff
@@ -375,7 +380,11 @@ export class SendScreenComponent extends React.Component<
                 <PrimaryCtaField
                     label={translate('App.labels.send')}
                     action={translate('App.labels.to')}
-                    value={formatAddress(this.state.toAddress, this.props.account.blockchain)}
+                    value={
+                        nameService
```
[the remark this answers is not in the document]
standreinmcp: I passed as nameService as an optional prop and then if it exists, immediately return it

src/core/utils/format-address.ts
```diff
@@ -1,7 +1,10 @@
 import { Blockchain } from '../blockchain/types';
 import { getBlockchain } from '../blockchain/blockchain-factory';
 
-export const formatAddress = (text: string, blockchain: Blockchain) => {
+export const formatAddress = (text: string, blockchain: Blockchain, nameService?: string) => {
+    if (nameService !== '') {
```
krisboit, line 5: we can be smarter :) 
```ts
if (text.match(/.+\.[a-zA-Z]{2,}$/gi)) return text;
```
so you don't really need an extra parameter

krisboit (changes requested):

krisboit (approved):

georgeburduhos (approved):

PR #931, opened 2020-09-17
New Summary Design

NovacAndrei:


src/core/utils/stats-get-value.ts
```diff
@@ -0,0 +1,29 @@
+import BigNumber from 'bignumber.js';
+import { getTokenConfig } from '../../redux/tokens/static-selectors';
```
georgeburduhos, line 2: pass the token config through params - we want to remove getTokenConfig from core

georgeburduhos (approved):

krisboit (approved):

PR #938, opened 2020-09-21
Zns implementation

standreinmcp:


src/core/blockchain/zilliqa/name-service.ts
```diff
@@ -21,15 +22,31 @@ export class NameService implements IBlockchainNameService {
...
         }
     }
 
-    public resolveName(text: string): Promise<IResolveNameResponse> {
-        return Promise.resolve({
-            address: text
-        });
+    public async resolveName(name: string): Promise<IResolveNameResponse> {
+        const node = namehash(name);
+        const znsToBech32 = fromBech32Address(ZNS);
```
georgeburduhos, line 43: the const name is toBech but you use fromBech :)

georgeburduhos (approved):

src/core/blockchain/zilliqa/name-service.ts
standreinmcp: Ah right, will change it

src/core/blockchain/types/name-service.ts
```diff
@@ -23,6 +25,6 @@ export interface IResolveTextResponse {
 }
 
 export interface IBlockchainNameService {
-    resolveText(text: string): Promise<IResolveTextResponse>;
-    resolveName(text: string): Promise<IResolveNameResponse>;
+    resolveText(blockchain: Blockchain, text: string): Promise<IResolveTextResponse>;
```
krisboit, line 28: no need for blockchain param

src/core/utils/format-address.ts
```diff
@@ -2,6 +2,7 @@ import { Blockchain } from '../blockchain/types';
 import { getBlockchain } from '../blockchain/blockchain-factory';
 
 export const formatAddress = (text: string, blockchain: Blockchain) => {
+    if (text === undefined) return '';
```
krisboit, line 5: it's simplier to have:
```ts
if (!text) return '';
```

src/core/utils/nameResolver.ts
```diff
@@ -0,0 +1,46 @@
...
+    ZIL = 'zil',
+    ETH = 'eth',
+    CRYPTO = 'crypto'
+}
+
+export const resolveNameForBlockchain = async (
+    blockchain: Blockchain,
+    name: string
+): Promise<{ address: string }> => {
+    const ending = name.split('.').pop();
```
krisboit, line 17: this need to be refactored, there is a smarter way to do it:
```ts
const ending = name.split('.').pop();
const { tld, service, record } = getBlockchain(blockchain).config.nameServices.find(
        item => typeof item.tld === 'string' ? item.tld === ending : !!ending.match(item.tld)
);

switch (service) {
    case BlockchainNameService.ZNS:
      return ethNameResolver(name, service, record);
      break; 
    case BlockchainNameService.ENS:
      return ethNameResolver(name, service);
      break;
    case BlockchainNameService.CNS:
      return ethNameResolver(name, service, record);
      break;
}
```

krisboit (changes requested):

src/core/blockchain/ethereum/name-service.ts
```diff
@@ -38,27 +46,19 @@ export class NameService implements IBlockchainNameService {
...
-            },
-            'latest'
-        ]);
-        const address =
-            '0x' + rawDecode(['address'], Buffer.from(data.result.replace('0x', ''), 'hex'))[0];
-        return Promise.resolve({ address });
+    public resolveName(name: string): Promise<IResolveNameResponse> {
+        const { mainNet } = this.client.network;
+        const ending = name.split('.').pop();
+        const { service, record } = config.nameServices.find(item =>
```
krisboit, line 52: this will crash if the result of find is `undefined`

src/core/blockchain/types/config.ts
```diff
@@ -4,10 +4,23 @@ import { ChainIdType } from '.';
 import { ITokenConfigState } from '../../../redux/tokens/state';
 import { IAffiliateBannerType } from '../../../components/affiliate-banner/types';
 
+export enum BlockchainNameService {
```
krisboit, line 7: these addresses are the same for testnets too ?

src/core/blockchain/ethereum/name-service.ts
```diff
@@ -38,27 +46,19 @@ export class NameService implements IBlockchainNameService {
...
+            typeof item.tld === 'string' ? item.tld === ending : !!ending.match(item.tld)
+        );
+        switch (service) {
```
krisboit, line 55: this method is basically the same with zilliqa one, move it in `GenericNameService`

krisboit (approved):

src/core/blockchain/ethereum/name-service.ts
standreinmcp: This is where the tests fail. I know it's duplicated, but this is what me and George came up with in order to solve it

src/core/blockchain/types/config.ts
standreinmcp: On the ziliqa blockchain: only testnet is currently supported. 
On the eth: the address works with mainnet and testnet aswel

PR #963, opened 2020-09-24
Stats redux

NovacAndrei:


src/redux/ui/stats/actions.ts
```diff
@@ -0,0 +1,47 @@
...
+    const state = getState();
+    const blockchain = account.blockchain;
+    const chainId = getChainId(state, blockchain).toString();
+
+    try {
+        const timestamp = getAccountStatsTimestamp(state, blockchain, chainId, account.address);
+
+        if (
+            !timestamp ||
+            moment().isAfter(moment(timestamp).add(FETCH_ACCOUNT_STATS_SECONDS, 'seconds'))
```
krisboit, line 26: you like moment a lot, but in some cases it's quite easy to use simple verifications. For instance in this case you could do this:
```
timestamp < Date.now() - FETCH_ACCOUNT_STATS_SECONDS * 1000
```

krisboit: you don't have to do the changes, i'm ok with your solution, it's just a hint for future :)

krisboit (approved):

georgeburduhos (approved):

PR #967, opened 2020-09-24
Fix account get balance (ZIL & NEAR)

NovacAndrei:


georgeburduhos (approved):

src/core/blockchain/zilliqa/client.ts
```diff
@@ -28,17 +30,15 @@ export class Client extends BlockchainGenericClient {
...
-                return Promise.resolve(new BigNumber(0));
-            }
-            return Promise.reject(result);
+            const data = await new ApiClient().validators.getBalance(
+                address,
+                Blockchain.ZILLIQA,
+                this.chainId.toString()
+            );
+
+            return data?.balance?.total || new BigNumber(0);
```
krisboit, line 39: we should return undefined here, and where we display the balances if the balance is undefined we should display `N/A` instead of `0`. When we display 0 users got scared they lost their tokens :)

krisboit: not necessary to implement it now, since we need to do it in many places. If it's simple do it, else add it in ZenHub.

krisboit (approved):

PR #970, opened 2020-09-24
Added copy to clipboard functionality

standreinmcp:


src/components/bottom-sheet/bottom-sheet.tsx
```diff
@@ -99,11 +99,12 @@ export class BottomSheetComponent extends React.Component<
                             activeOpacity={1}
                         />
                         <DashboardMenuBottomSheet
+                            selectedAccount={this.props.selectedAccount}
```
georgeburduhos, line 102: there is a selector getSelectedAccount - it returns the current account that you are on.
No need for external props but be careful cause selectedAccount will be undefined when the app is first installed

src/components/bottom-sheet/dashboard-menu-bottom-sheet/dashboard-menu-bottom-sheet.tsx
```diff
@@ -17,14 +17,17 @@ import { connect } from 'react-redux';
...
 
 interface IExternalProps {
     snapPoints: { initialSnap: number; bottomSheetHeight: number };
     onClose: () => void;
+    selectedAccount: IAccountState;
 }
 
 export interface IReduxProps {
     blockchain: Blockchain;
+    address: string;
```
georgeburduhos, line 30: you don’t use this anymore

georgeburduhos (changes requested):

src/components/bottom-sheet/dashboard-menu-bottom-sheet/styles.ts
```diff
@@ -35,6 +37,10 @@ export default (theme: ITheme) =>
         title: {
             lineHeight: normalizeFontAndLineHeight(21)
         },
+        subtitle: {
+            lineHeight: normalizeFontAndLineHeight(21),
+            fontSize: 11
```
NovacAndrei, line 42: Please use `fontSize: normalizeFontAndLineHeight (11)`

src/components/bottom-sheet/dashboard-menu-bottom-sheet/dashboard-menu-bottom-sheet.tsx
```diff
@@ -161,6 +178,16 @@ export class DashboardMenuBottomSheetComponent extends React.Component<
                             iconName: IconValues.FLASH_OFF,
                             onPress: () => this.connectedWebsites()
                         })}
+                    {Platform.OS !== 'web' &&
```
NovacAndrei, line 181: I think we also should add this functionality on web, it would be helpful.
HINT: make sure if you add this on web, it will be visible

krisboit: make sure it's visible and working

src/components/bottom-sheet/bottom-sheet.tsx
standreinmcp: I thought we try to not use redux in components. that's why I passed it as a prop

NovacAndrei (changes requested):

src/components/bottom-sheet/dashboard-menu-bottom-sheet/styles.ts
```diff
@@ -16,13 +16,15 @@ export default (theme: ITheme) =>
...
+            marginBottom: BASE_DIMENSION * 2,
+            alignItems: 'center'
         },
         iconContainer: {
             backgroundColor: theme.colors.appBackground,
             padding: BASE_DIMENSION,
             marginRight: BASE_DIMENSION * 2,
-            borderRadius: BORDER_RADIUS
+            borderRadius: BORDER_RADIUS,
+            maxHeight: BASE_DIMENSION * 6
```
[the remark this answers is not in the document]
krisboit: @NovacAndrei is there a problem with maxHeight ?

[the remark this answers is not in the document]
NovacAndrei: Not exactly with maxHeight, but icons do not always have the same sizes (width and height) and because of that the parent views does not always look the same.

Look this for example:
image

If we enforce a width and height, remove paddings and aligns everything center, we shouldn't have any problems. On this icons we should not depend on paddings.

krisboit (approved):
@NovacAndrei you have the final word :)

NovacAndrei (approved):

NovacAndrei:
All good now 👍

georgeburduhos (approved):

PR #994, opened 2020-09-30
General: Fixed quick stake amount on quick stake banner

standreinmcp:
Verifying now on the right amount, so the banner appears when you have over 1000 zil and dissapears when below 1000 zil.


src/components/quick-delegate-banner/quick-delegate-banner.tsx
```diff
@@ -31,7 +33,7 @@ export const QuickDelegateBannerComponent = (
 
     const blockchainConfig = getBlockchain(props.blockchain);
 
-    if (!props.account) {
+    if (!props.account && !props.accountStats.chartStats) {
```
NovacAndrei, line 36: Just for safety, use `if (!props.account && !props.accountStats?.chartStats) {`

georgeburduhos (commented):
What happens if we decide to remove Available from chart or move it to another place inside the array?

standreinmcp:
@georgeburduhos then other functionalities will need to be changed too: for ex: I believe Available amount in Summary is implemented just the same as this one.

src/components/quick-delegate-banner/quick-delegate-banner.tsx
```diff
@@ -48,27 +50,28 @@ export const QuickDelegateBannerComponent = (
...
 
             const minAmount = new BigNumber(clientMinAmount).plus(new BigNumber(fees));
-
             // getMinimumAmountDelegate + fees + 1%
             setMinimumAmountDelegate(minAmount.plus(minAmount.dividedBy(100)));
         })();
     }, [props.blockchain, props.account.address, props.chainId]);
 
     if (tokenConfig.ui.tokenScreenComponent === TokenScreenComponentType.DELEGATE) {
+        const avaialableChartData = props.accountStats?.chartStats.find(
```
krisboit, line 64: don't we have this value in redux ? I'm not happy with this solution, if we will add internationalisation `item.title` will not be `Available`. This data should be use only for display purposes.

src/core/blockchain/near/client.ts
```diff
@@ -29,17 +30,22 @@ export class Client extends BlockchainGenericClient {
...
         try {
             const data = await new ApiClient().validators.getBalance(
                 address,
                 Blockchain.NEAR,
                 this.chainId.toString()
             );
 
-            return data?.balance?.total || new BigNumber(0);
+            return (
+                { total: data?.balance?.total, available: data?.balance?.available } || {
```
krisboit, line 42: `{} || ..` an empty object is not a falsy value... so `|| {..}` is useless

src/redux/wallets/reducer.ts
```diff
@@ -26,10 +26,12 @@ import { RESET_ALL_DATA, EXTENSION_UPDATE_STATE } from '../app/actions';
 const intialState: IWalletsState = {};
 
 const newBalance = (oldBalance: any, action: any) => ({
-    value: action.data.balance ? action.data.balance : oldBalance?.value,
+    value: action.data.balance ? action.data.balance.total : oldBalance?.value,
```
krisboit, line 29: `value` should be `balance.available`

krisboit (changes requested):

src/redux/wallets/reducer.ts
standreinmcp: I added available 2 lines after this. so right now we have: available, total and value. Value should still be available ? 
available: action.data.balance ? action.data.balance.available : oldBalance?.available,

krisboit: balance.value is used in many places in our app, and in most of the cases that value represents the available amount from user's account. Here you set the total amount on `value` which will lead to misleading information.

georgeburduhos: If we change the `value` we must handle state migration as well
My suggestion for now: keep `value` as available and add `total` to use it where we need
@krisboit

georgeburduhos (approved):

src/core/blockchain/types/client.ts
```diff
@@ -11,6 +11,11 @@ export interface IBlockInfo {
     hash?: string;
 }
 
+export interface IBalance {
+    total: BigNumber;
```
krisboit, line 15: There are 2 options here:
1. keep BigNumber as type here, but you have to be sure that BigNumber instances don't end up directly in redux. We would have problems on state rehydrate.
2. transform to string, redux will be fine, use BigNumber only where we need to use the amount to calculate something

src/core/blockchain/near/client.ts
```diff
@@ -29,17 +30,22 @@ export class Client extends BlockchainGenericClient {
...
+    public async getBalance(address: string): Promise<IBalance> {
         try {
             const data = await new ApiClient().validators.getBalance(
                 address,
                 Blockchain.NEAR,
                 this.chainId.toString()
             );
-
-            return data?.balance?.total || new BigNumber(0);
+            if (data.balance.total !== undefined && data.balance.available !== undefined)
```
krisboit, line 40: you don't really need an if here, you can do something like:
```ts
return {
  total: data?.balance?.total || 0
  available: ...
}
```

krisboit: the types may not be ok in my example

src/core/blockchain/types/client.ts
```diff
@@ -31,7 +36,7 @@ export abstract class BlockchainGenericClient {
         this.http = new HttpClient(url);
     }
 
-    public abstract getBalance(address: string): Promise<BigNumber>;
+    public abstract getBalance(address: string): Promise<IBalance | BigNumber>;
```
krisboit, line 39: I'm not ok with the `IBalance | BigNumber`, it should be one or the other.

src/components/quick-delegate-banner/quick-delegate-banner.tsx
```diff
@@ -31,7 +33,7 @@ export const QuickDelegateBannerComponent = (
 
     const blockchainConfig = getBlockchain(props.blockchain);
 
-    if (!props.account) {
+    if (!props.account && !props.accountStats?.chartStats) {
```
krisboit, line 36: is there any dependency on chartStats for this banner ?

krisboit (changes requested):

src/components/quick-delegate-banner/quick-delegate-banner.tsx
standreinmcp: not anymore

src/core/blockchain/types/client.ts
standreinmcp: Before, value was as a BigNumber, I don't think it gets to redux as a BigNumber but I'll double check

src/components/quick-delegate-banner/quick-delegate-banner.tsx
krisboit: ok, remove it :)

src/core/blockchain/types/client.ts
krisboit: please do

standreinmcp: It's fine, they are all converted to string before

src/core/blockchain/ethereum/client.ts
```diff
@@ -24,9 +25,9 @@ export class Client extends BlockchainGenericClient {
         this.utils = new ClientUtils(this);
     }
 
-    public getBalance(address: string): Promise<BigNumber> {
+    public getBalance(address: string): Promise<IBalance> {
         return this.http.jsonRpc('eth_getBalance', [fixEthAddress(address), 'latest']).then(res => {
-            return new BigNumber(res.result, 16);
+            return { total: new BigNumber(res.result, 16), available: new BigNumber(0) };
```
krisboit, line 30: available shoudn't be 0. On ETH total = available

krisboit (changes requested):

krisboit (approved):

PR #1002, opened 2020-10-01
Fixed Near double name appearing when entering address.

standreinmcp:


src/screens/send/components/add-address/add-address.tsx
```diff
@@ -166,7 +166,7 @@ export class AddAddressComponent extends React.Component<
             const response = await blockchainInstance
                 .getClient(this.props.chainId)
                 .nameService.resolveText(text);
-            this.setState({ resolvedAddress: response.address });
+            if (text !== response.address) this.setState({ resolvedAddress: response.address });
```
krisboit, line 169: is this the only place, don't we have an address field on send screen also ?

standreinmcp: This fixes everything because if it matches, the resolvedAddress will be ''. In the send.tsx we only get the right address, the other one is empty string. because we don't set the state

NovacAndrei:
Please double check on NEAR `Step 3: Confirm transaction

krisboit (approved):

NovacAndrei (approved):

PR #1007, opened 2020-10-01
Fixed validators tab empty state. Now showing an image when the are n…

standreinmcp:
…o validators


src/screens/token/components/delegate-token/components/tabs/validators-tab/validators-tab.tsx
```diff
@@ -106,8 +105,12 @@ export class ValidatorsTabComponent extends React.Component<
                         />
                     </View>
                     {this.state.validatorsFilteredList.length === 0 ? (
-                        <View style={{ flex: 1 }}>
-                            <LoadingIndicator />
+                        <View style={styles.emptySection}>
+                            <Image
```
NovacAndrei, line 109: Here you should have all the 3 states:

1. Loading
2. Empty state
3. Validators list

We can have a quick call to explain

src/redux/ui/delegated-validators/selectors.ts
```diff
@@ -11,6 +11,6 @@ export const getDelegatedValidators = (
         (state.ui.delegatedValidators &&
             state.ui.delegatedValidators[blockchain] &&
             state.ui.delegatedValidators[blockchain][chainId as string]) ||
-        []
+        undefined
```
krisboit, line 14: no need for `|| undefined`. What you are actually doing is : if the value is undefined return undefined.

krisboit (approved):

georgeburduhos (approved):

PR #1019, opened 2020-10-02
Implemented account verification based on the network

standreinmcp:
Now you can see your account only on the network that you've made that account. Accounts with the network undefined will be available on both networks


src/redux/wallets/selectors.ts
```diff
@@ -115,9 +115,13 @@ export const getSelectedAccountTransactions = (state: IReduxState): IBlockchainT
...
 
 export const getAccounts = (state: IReduxState, blockchain: Blockchain): IAccountState[] => {
-    const accounts = getSelectedWallet(state).accounts.filter(acc => {
-        if (acc.blockchain === blockchain) {
-            return acc;
+    const chainId = getChainId(state, blockchain);
+    const accounts = getSelectedWallet(state).accounts.filter(account => {
+        if (
+            account.blockchain === blockchain &&
+            (chainId === account?.chainId || account?.chainId === undefined)
```
krisboit, line 122: as a hint, you don't need to always use `left?.right`. If `left` is always an object  `left.right` will evaluate to `undefined` if `right` is not defined.

krisboit (approved):

NovacAndrei (approved):

PR #1052, opened 2020-10-07
Implemented public beta badge on the dashboard

standreinmcp:


src/screens/dashboard/dashboard.tsx
```diff
@@ -452,6 +453,7 @@ export class DashboardScreenComponent extends React.Component<
 
         return (
             <View testID="dashboard-screen" style={[styles.container, { height: containerHeight }]}>
+                <StakingBetaBadge />
```
krisboit, line 456: we should add a throttle for this, so we will able to display and hide without a app release. Talk with George about this, it's very simple :)

krisboit (approved):

src/core/utils/remote-feature-config.ts
```diff
@@ -65,7 +71,9 @@ export const isFeatureActive = (feature: RemoteFeature): boolean => {
     }
 
     if (
-        (feature === RemoteFeature.NEAR ||
+        (feature === RemoteFeature.BETA_BADGE ||
+            feature === RemoteFeature.ZIL ||
```
georgeburduhos, line 75: Check my branch. - there is a thing to consider.
If in the list that comes from firebase you see * then the feature will be active.
This will help us show to all users specific users

krisboit (approved):

src/core/utils/remote-feature-config.ts
```diff
@@ -64,16 +70,13 @@ export const isFeatureActive = (feature: RemoteFeature): boolean => {
...
-            feature === RemoteFeature.COSMOS ||
-            feature === RemoteFeature.CELO ||
-            feature === RemoteFeature.DEV_TOOLS) &&
-        featuresConfig
-    ) {
+    if (featuresConfig) {
         const values = featuresConfig[feature] || [];
         if (values.length > 0) {
-            return values.indexOf(DeviceInfo.getUniqueId()) >= 0;
+            return (
```
krisboit, line 76: you need to add one more check here, `Array.isArray(values)` `values.length` will fail if values is a number for example

krisboit (changes requested):

krisboit (approved):

georgeburduhos (approved):

PR #1058, opened 2020-10-07
ledger flow

georgeburduhos:


src/core/utils/format-string.ts
```diff
@@ -21,3 +21,10 @@ export const valuePrimaryCtaField = (validators: IValidator[]): string => {
 export const formatValidatorName = (str: string, length: number): string => {
     return str.length > length ? str.slice(0, length) + '...' : str;
 };
+
+export const formatPlural = (value: number): string => {
+    if (value === 1) return '1st';
```
krisboit, line 26: Oh no :) will this work when we add other languages ? :)
The best approach here is to use our translation system, that way you will cover all cases. Maybe in this case is not applicable, but for instance you function will not handle correct 21, 22, 101, 102, 103... and many others. Our translation system is very capable. So you just have to add this in our en.ts file:

```ts
"ParentKey": {
   "ledgerSignButton": {
      ordinal: true,
      text: "Sign {{txNumber}}th transaction",
      forms: {
         one: "Sign {{txNumber}}st transaction",
         two: "Sign {{txNumber}}nd transaction",
         few: "Sign {{txNumber}}rd transaction",
         other: "Sign {{txNumber}}th transaction", // this is not needed since we have the default text above
    }
}
```
Or you can have only the numbers in a key and have the button text as "SIGN {{number}} transaction".

This way the ordinal numbers formatted correctly, like in this table https://linguapress.com/grammar/ordinals.htm and fully customisable for any language.

src/redux/wallets/actions/pos-actions.ts
```diff
@@ -286,9 +274,29 @@ export const posAction = (
...
+};
+
+export const signAndSendTransactions = (transactions: IBlockchainTransaction[]) => async (
+    dispatch: Dispatch<IAction<any>>,
+    getState: () => IReduxState
+) => {
+    const state = getState();
 
+    const appWallet = getSelectedWallet(state);
+    const account = getSelectedAccount(state); // Not sure its ok...
```
krisboit, line 290: you can do this, it's safe, or another option would be to get the account from the first tx from the list, I think it's safe to assume that all transactions are made from the same account.

src/screens/pos-actions/process-transactions/process-transactions.tsx
```diff
@@ -387,6 +387,97 @@ export class ProcessTransactionsComponent extends React.Component<
...
+                        primary
+                        onPress={() =>
+                            this.onPressSignTransaction(
+                                this.props.transactions[this.state.currentIndex]
+                            )
+                        }
+                        wrapperStyle={styles.continueButton}
+                        disabled={false}
+                    >
+                        {`${translate('App.labels.sign')} ${formatPlural(
```
krisboit, line 461: this approach is wrong, Not all languages have the same order of words, for example `Sign 1st transaction` in german is `Erste Transaktion unterschreiben`. 
PS: Erste = 1st :)

krisboit (approved):

NovacAndrei (approved):

PR #1066, opened 2020-10-08
Implemented minimum amount to keep in account when staking

standreinmcp:


src/core/blockchain/types/config.ts
```diff
@@ -102,4 +102,5 @@ export interface IBlockchainConfig {
     };
     defaultOrder: number;
     nameServices?: INameService[];
+    amountToKeepInAccount?: BigNumber;
```
georgeburduhos, line 105: not optional - put values in every blockchain
We need to avoid as much as possible optional values in config even if it means to return a 0 value

georgeburduhos (changes requested):

src/core/blockchain/near/stats.ts
```diff
@@ -30,6 +30,7 @@ export class Stats extends GenericStats<Client> {
 
         return new BigNumber(data.balance.available)
```
krisboit, line 31: this is not entirely correct. You have the following case:
- amount to keep = 100
- user has available balance 50
- user has unstaked balance 500
The user should be able to stake 500. Your version will return 450.

Basically you have to do this:
```
var aviableToDelegate = unstakedBalance
if (availableBalance > amountToKeep) aviableToDelegate += availableBalance - amountToKeep
```

src/core/blockchain/zilliqa/config.ts
```diff
@@ -172,7 +172,8 @@ export const config: IBlockchainConfig = {
             service: BlockchainNameService.CNS,
             record: ['crypto.ZIL.address']
         }
-    ]
+    ],
+    amountToKeepInAccount: new BigNumber(Math.pow(10, 14))
```
krisboit, line 176: JS is not reliable for big numbers, Math.pow uses native js numbers so it's not safe. use `BigNumber.pow` or use a string value.

krisboit (changes requested):

src/core/blockchain/zilliqa/stats.ts
```diff
@@ -24,6 +25,12 @@ export class Stats extends GenericStats<Client> {
             this.client.chainId.toString()
         );
 
-        return new BigNumber(data.balance.available).toFixed();
+        let availableToDelegate = new BigNumber(data.balance.unstaked);
+        if (availableToDelegate > config.amountToKeepInAccount) {
```
krisboit, line 29: i thinik you should check this instead `data.balance.available > config.amountToKeepInAccount` also you cannot compare bignumbers like this. use lt gt functions

src/core/blockchain/near/stats.ts
```diff
@@ -28,8 +28,12 @@ export class Stats extends GenericStats<Client> {
             validatorId
         );
 
-        return new BigNumber(data.balance.available)
-            .plus(new BigNumber(data.balance.unstaked))
-            .toFixed();
+        let availableToDelegate = new BigNumber(data.balance.unstaked);
+        if (availableToDelegate > config.amountToKeepInAccount) {
```
krisboit, line 32: same here.

krisboit (changes requested):

georgeburduhos (approved):

krisboit (approved):

PR #1068, opened 2020-10-08
Update prod endpoints

NovacAndrei:


src/core/address-monitor/index.ts
```diff
@@ -5,7 +5,7 @@ import { IWalletsState, IAccountState } from '../../redux/wallets/state';
 import { CONFIG } from '../../config';
 import { HttpClient } from '../utils/http-client';
 
-const url = CONFIG.dataApiUrl + '/notifications/register';
+const url = CONFIG.walletApiBaseUrl + '/data//notifications/register';
```
krisboit, line 8: the url path seems broken you have 2 `//`.
also, is this still used ? '/data/notifications/register' was the old api right ?

krisboit (changes requested):

krisboit (approved):

Issue #1037, opened 2020-10-06
[ZIL] All moonlet operation send a tiny little amount of ZIL 2423 even when the transition does not accept $ZIL


i3uRi:
All moonlet operation send a tiny little amount of ZIL 2423 even when the transition does not accept $ZIL. It should be zero.

![image.png](https://images.zenhubusercontent.com/5b72918402b9fd36cfbf71b4/363a663f-c31f-4329-b16f-926558bf6adc)

PR #1074, opened 2020-10-08
[ZIL] All moonlet operation send a tiny little amount of ZIL 2423 eve…

georgeburduhos:
…n when the transition does not accept $ZIL #1037

### DIFF
--- a/src/core/wallet/hw-wallet/ledger/apps/zil.ts
+++ b/src/core/wallet/hw-wallet/ledger/apps/zil.ts
@@ -39,7 +39,7 @@ export class Zil {
                 .fromBech32Address(tx.toAddress)
                 .replace('0x', '')
                 .toLowerCase(),
-            amount: tx.amount.toString(),
+            amount: tx.amount ? tx.amount.toString() : '0',
             pubKey: tx.publicKey,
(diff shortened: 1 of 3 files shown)


src/core/wallet/hw-wallet/ledger/apps/zil.ts
```diff
@@ -39,7 +39,7 @@ export class Zil {
                 .fromBech32Address(tx.toAddress)
                 .replace('0x', '')
                 .toLowerCase(),
-            amount: tx.amount.toString(),
+            amount: tx.amount ? tx.amount.toString() : '0',
```
krisboit, line 42: if amount is BigNumber we need to use `toFixed()` or do a new BigNumber(tx.amount).toFixed() to be sure

krisboit: on zilliqa we have fewer decimals but if someone sends a big amount it will be something like: `1.0e+15`

krisboit (changes requested):

NovacAndrei (approved):

krisboit (approved):

PR #1079, opened 2020-10-08
Fixed amount on transaction details for Rewards

standreinmcp:


src/core/blockchain/zilliqa/contracts/staking.ts
```diff
@@ -344,7 +346,7 @@ export class Staking {
 
         transaction.data = {
             method: 'Claim Rewards',
-            params: [validator.id, undefined],
+            params: [validator.id, chartData.data.value],
```
krisboit, line 349: i don't know if this is a good idea, this impacts the transaction itself, so for reward we may need to check transaction logs after it was executed, to get the value.

krisboit (changes requested):

src/core/blockchain/zilliqa/contracts/staking.ts
georgeburduhos: @krisboit Its ok, that data is used only by us internally, is not send to the sign, broadcast or blockchain

georgeburduhos: My issue here is that the value is taken from chartData - but i think for this iteration is ok.
@standreinmcp  - add a todo comment here to not forget to find a better solution

krisboit (approved):

georgeburduhos (approved):

PR #1160, opened 2020-10-11
 Remove -&+ button from Stake flow

NovacAndrei:


src/screens/pos-actions/pos-action-delegate/delegate-select-validator/delegate-select-validator.tsx
```diff
@@ -156,71 +152,73 @@ export class DelegateSelectValidatorComponent extends React.Component<
...
-            blockchainInstance.config.ui.validator.maximumNumberOfValidators <=
-                this.state.nrValidators;
+        // const { styles } = this.props;
+        // const blockchainInstance = getBlockchain(this.props.blockchain);
+        // const maximumNumberOfValidatorsReached =
+        //     blockchainInstance.config.ui.validator.maximumNumberOfValidators &&
+        //     blockchainInstance.config.ui.validator.maximumNumberOfValidators <=
+        //         this.state.nrValidators;
+
+        // Hide this feature for now
```
georgeburduhos, line 162: You need to add the entire list of validators here and on redelegate as well as is done on quick stake - now from what i remember it displays only the first validator

georgeburduhos (changes requested):

georgeburduhos (approved):

PR #1175, opened 2020-10-12
[All] Added getBalance request on account-tab

standreinmcp:


src/screens/token/components/delegate-token/components/tabs/account-tab/account-tab.tsx
```diff
@@ -78,6 +85,18 @@ export class AccountTabComponent extends React.Component<
         this.props.fetchValidators(this.props.account, PosBasicActionType.DELEGATE);
         this.props.fetchDelegatedValidators(this.props.account);
         this.props.fetchAccountDelegateStats(this.props.account, this.props.token);
+        this.onFocus();
+    }
+
+    public onFocus() {
```
NovacAndrei, line 91: Can change this to `getBalance()` and cover one more case: when navigate back from `Quick stake` should also call this.
Search in the project about focus. You'll find an example

src/redux/wallets/reducer.ts
```diff
@@ -27,12 +27,18 @@ const intialState: IWalletsState = {};
 
 const newBalance = (oldBalance: any, action: any) => ({
     // TODO migration and change value to actual avaialble and total values
-    value: action.data.balance ? action.data.balance.available : oldBalance?.value,
+    value: action.data.balance
```
georgeburduhos, line 30: why this changes?

PR #1204, opened 2020-10-15
[NEAR] update amountToKeepInAccount for lockup

NovacAndrei:


georgeburduhos (approved):

src/core/blockchain/zilliqa/stats.ts
```diff
@@ -26,9 +26,11 @@ export class Stats extends GenericStats<Client> {
         );
 
         let availableToDelegate = new BigNumber(0);
-        if (new BigNumber(data.balance.available).gt(config.amountToKeepInAccount)) {
```
krisboit, line 29: we need to be sure account.type is always defined, otherwise this will fail

NovacAndrei: We have this state migration in place

image

PR #1219, opened 2020-10-20
 Amount threshold message

standreinmcp:


src/core/blockchain/celo/config.ts
```diff
@@ -151,6 +151,9 @@ export const config: IBlockchainConfig = {
     defaultOrder: 1,
     amountToKeepInAccount: {
         [AccountType.DEFAULT]: new BigNumber(0)
+    },
+    amountToKeepForTheFees: {
```
krisboit, line 155: why are we adding this ? `amountToKeepInAccount` is exactly for fees :)

src/screens/send/components/enter-amount/enter-amount.tsx
```diff
@@ -46,6 +46,10 @@ export const EnterAmountComponent = (
         onChange(addedAmount.toFixed());
     };
 
+    const verifyInsuficientFunds = () => {
+        if (new BigNumber(availableAmount).eq(0)) return true;
```
krisboit, line 50: is this necessary ? if yes it seems we have a bug when we calculate the amounts and compute `insufficientFunds`

src/screens/validator-screen/validator-screen.tsx
```diff
@@ -56,11 +77,39 @@ const navigationOptions = ({ navigation }: any) => ({
...
+
+    async getBalance() {
+        const { total } = await getBlockchain(this.props.blockchain)
+            .getClient(this.props.chainId)
+            .getBalance(this.props.account.address);
+        this.setState({ total });
+    }
+
+    handleOnClaimAlert() {
+        const config = getBlockchain(this.props.blockchain).config;
```
krisboit, line 106: i think we can move this in can perform action, and just return an error code. Based on the error code we can display the texts in the warning popup

krisboit (changes requested):

src/components/cta-group/cta-group.tsx
```diff
@@ -69,6 +70,20 @@ export const CtaGroupComponent = (
                                 key={`cta-${index}`}
                                 leftIcon={cta.iconName}
                                 onPress={() => {
+                                    if (
+                                        props.params.onClaimAlert &&
+                                        cta.navigateTo.params.basicAction ===
```
NovacAndrei, line 75: You have `PosBasicActionType.CLAIM_REWARD_NO_INPUT`

src/core/i18n/translation/locales/en.ts
```diff
@@ -95,7 +95,7 @@ export const translation: ITranslations = {
                 retry: 'Retry',
                 reset: 'Reset',
                 resetAll: 'Reset all',
-                restake: 'Restake',
+                restake: 'Switch Node',
```
NovacAndrei, line 98: We should not do this, if you need `App.labels.restake` in another context, you'll get `Switch Node` instead

NovacAndrei (changes requested):

src/screens/send/components/enter-amount/enter-amount.tsx
standreinmcp: insuficientFunds calculation is done only on input. I can make it to fire on componentDidMount or the current implementation

src/core/blockchain/celo/config.ts
standreinmcp: This task: https://app.zenhub.com/workspaces/moonlet-wallet-5d93018da1bdb00001e54a59/issues/moonlet/wallet-app/1206
 says to check for 1 zil. The amountToKeepInAccount is 100 zil.

standreinmcp: I should probably rename it

src/core/i18n/translation/locales/en.ts
standreinmcp: We only use App.labels.restake in one instance: zilliqa config. But I can modify the name if you want

NovacAndrei: Yes, we should keep the key value pair from en.ts self explanatory.

src/core/i18n/translation/locales/en.ts
```diff
@@ -577,10 +577,12 @@ export const translation: ITranslations = {
             selectStakingPool: 'Selecting staking pool',
             unselectStakingPool: 'Unselecting staking pool',
             allBalanceNotice:
-                'Notice: Moonlet keeps some tokens in your account for future transaction fees.',
+                'Notice: Moonlet keeps a minimum of 100 ZIL in your account for future transactions like Stake, Claim, Unstake or Switch Node.',
```
krisboit, line 580: this message is shown on all blockchains, so with this change we will diplay 100 ZIL on Near also. If we want this we need to use parameters in transalation

krisboit (changes requested):

krisboit (approved):

NovacAndrei (approved):

PR #1220, opened 2020-10-20
[NEAR] Fix unstaked balance

NovacAndrei:


krisboit (approved):

src/screens/pos-actions/pos-action-quick-delegate/quick-delegate-enter-amount/quick-delegate-enter-amount.tsx
```diff
@@ -87,9 +87,15 @@ export class QuickDelegateEnterAmountComponent extends React.Component<
         const tokenConfig = getTokenConfig(this.props.blockchain, this.props.token.symbol);
         try {
             this.setState({ loading: true });
+
+            let validator;
+            if (this.props.validators.length === 1) {
```
georgeburduhos, line 92: What happens if the number of selected validators is more then 1? Or is not possible?

georgeburduhos (approved):

PR #1236, opened 2020-10-28
Smart Screen & Widgets

NovacAndrei:
Remove `redux/screens/stats`


src/screens/smart-screen/smart-screen.tsx
```diff
@@ -0,0 +1,200 @@
...
+        };
+    }
+
+    public componentDidMount() {
+        this.props.fetchScreenData(this.props.context);
+    }
+
+    public componentDidUpdate(prevProps: IReduxProps & IExternalProps) {
+        if (
+            (this.props.account && this.props.account !== prevProps.account) ||
```
georgeburduhos, line 89: make a hash here and compare it. its easier. check `ZilliqaTransactionUpdateComponent`  did update function

georgeburduhos (approved):

src/screens/smart-screen/smart-screen.tsx
```diff
@@ -0,0 +1,216 @@
...
+            (this.props.chainId && this.props.chainId !== prevProps.chainId) ||
+            this.props.context?.tab !== prevProps.context?.tab
+        ) {
+            this.props.fetchScreenData(this.props.context);
+            this.getScreenData();
+        }
+
+        // TODO: needs to be fixed for tab account
+
+        if (
```
krisboit, line 103: this is not needed, just use props directly, loose screen data from state

src/screens/smart-screen/smart-screen.tsx
```diff
@@ -0,0 +1,216 @@
...
+                error: undefined
+            });
+        }
+    }
+
+    public render() {
+        const { styles, theme } = this.props;
+        const { screenData } = this.state;
+
+        if (this.state.isLoading) {
```
krisboit, line 153: you should display a skeleton only if you the state is loading and there is no data in screen data

src/screens/smart-screen/smart-screen.tsx
```diff
@@ -0,0 +1,216 @@
...
+                            withdraw: this.props.withdraw
+                        }}
+                        account={this.props.account}
+                        chainId={this.props.chainId}
+                    />
+                </View>
+            );
+        }
+
+        if (this.state.error) {
```
krisboit, line 192: We have to decide if we should display error if we have something in screen data, yes the data could be old but maybe it's better to display data instead of an error... ask @i3uRi about this...

src/screens/smart-screen/smart-screen.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+interface IExternalProps {
+    context: IScreenContext;
+}
+
+const mapStateToProps = (state: IReduxState) => {
+    const account = getSelectedAccount(state);
+    const chainId = account && getChainId(state, account.blockchain);
+
+    return {
```
krisboit, line 31: well this is not scalable at all :) let's do something generic. you should return here a `screenData` key that has the value of `state.ui.screens.data[ownProps.context.screen]` this way you return only what you need in the screen. Of course you need to handle some errors cases.

krisboit: this way you can remove `getScreenData` method :) i think it won't be needed anymore

src/screens/pos-actions/process-transactions/process-transactions.tsx
```diff
@@ -229,13 +229,15 @@ export class ProcessTransactionsComponent extends React.Component<
...
-                        ' ' +
-                        tx.additionalInfo?.validatorName ||
+                const validatorName =
+                    tx.additionalInfo?.validatorName ||
                     tx.additionalInfo?.validator?.name ||
                     tx.additionalInfo?.validatorId ||
                     tx.additionalInfo?.validator?.id;
+
+                if (validatorName) {
+                    translate('App.labels.from').toLowerCase() + ' ' + validatorName;
```
krisboit, line 239: this line doesn't do anything ...

src/redux/wallets/actions/pos-actions.ts
```diff
@@ -503,3 +486,21 @@ const transactionsBroadcasted = (txs: IBlockchainTransaction[]): boolean => {
         }).length > 0
     );
 };
+
+export const buildDummyValidator = (id: string, name?: string): IValidator => {
```
krisboit, line 490: why do we need this ?

src/redux/ui/screens/data/actions.ts
```diff
@@ -0,0 +1,83 @@
...
+
+            blockchain: account.blockchain,
+            chainId: String(chainId),
+            address: account.address,
+            accountType: AccountType.DEFAULT
+        }
+    };
+
+    try {
+        const screenResponse = await apiClient.http.post('/walletUi/screen', body);
```
krisboit, line 60: you should update the state before doing the request, switch `isLoading` to `true` and assign `undefined` to `error` field

src/redux/ui/screens/data/actions.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    const apiClient = new ApiClient();
+
+    const body: IScreenRequest = {
+        context: {
+            screen: context.screen,
+            tab: context?.tab
+        },
+        user: {
+            os: Platform.OS as 'ios' | 'android',
```
krisboit, line 41: for extension we will have 'web' here

src/redux/ui/screens/data/actions.ts
```diff
@@ -0,0 +1,83 @@
...
+            deviceId: state.preferences.deviceId,
+            theme: 'dark',
+            country: 'Moon',
```
krisboit, line 44: make this optional till we will add a relevant value :)

src/components/widgets/utils.tsx
```diff
@@ -0,0 +1,154 @@
...
+const amountFromStd = (value: BigNumber | number | string, decimals: number): BigNumber => {
+    return new BigNumber(value).dividedBy(new BigNumber(10).pow(decimals));
+};
+
+const formatCurrencyData = (data: ICurrencyData): string => {
+    let text = '';
+    let value = data.value;
+
+    if (data?.decimals) {
+        value = amountFromStd(value, data.decimals).toFixed();
```
krisboit, line 46: i wouldn't call `.toFixed()` here. let it as a big number.

src/components/widgets/utils.tsx
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    text = formatNumber(Number(value), {
```
krisboit, line 49: before calling `formatNumber` you should do the rounding, with the specific flag (ROUND_UP or ROUND_DOWN). 
then in formatNumber you could use `value.toNumber()`

krisboit (changes requested):

krisboit (approved):

PR #1290, opened 2020-11-04
Widget wrapper modules

NovacAndrei:


src/components/widgets/components/module-wrapper/state-modifiers/quick-stake-select-validator.ts
```diff
@@ -0,0 +1,27 @@
+import { getChainId } from '../../../../../redux/preferences/selectors';
+import { IReduxState } from '../../../../../redux/state';
+import { getScreenDataKey } from '../../../../../redux/ui/screens/data/reducer';
+import { getSelectedAccount, getSelectedWallet } from '../../../../../redux/wallets/selectors';
+import { IScreenModule } from '../../../types';
+
+export const quickStakeSelectValidator = (state: IReduxState, module: IScreenModule): string => {
```
krisboit, line 7: rename to `quickStakeSelectedValidator`

src/components/widgets/components/two-lines-text-banner/two-lines-text-banner.tsx
```diff
@@ -25,46 +25,55 @@ const TwoLinesStakeBannerComponent = (
...
             </View>
-        </TouchableOpacity>
+
+            {data?.icon?.value && (
+                <Icon name={data.icon.value} size={normalize(50)} style={styles.icon} />
+            )}
+        </View>
     );
+
+    if (module?.cta) {
```
krisboit, line 51: for now it's fine, but we need a function to handle cta objects from API in a generic way. This is a generic component but cta functionality is limited

src/screens/quick-stake-select-validator/quick-stake-select-validator.tsx
```diff
@@ -0,0 +1,141 @@
...
+        super(props);
+        this.state = {
+            selectedValidators: []
+        };
+    }
+
+    public componentDidUpdate(prevProps: IReduxProps) {
+        if (this.props.validators && this.props.validators !== prevProps.validators) {
+            const selectedValidators = [];
+            for (const v of this.props.validators) {
```
krisboit, line 81: move this for inside `onPress` function from bottom cta component, this way you don't need local state anymore, less work less reders ;)

src/screens/smart-screen/smart-screen.tsx
```diff
@@ -168,7 +189,18 @@ export class SmartScreenComponent extends React.Component<IReduxProps & IExterna
 
         if (screenData) {
             if (this.state.loadingScreenData) {
```
krisboit, line 191: it's ok for now, but is not quite scalable, maybe we should have a param of function type `getLoadingSkeleton` so we can customize the skeleton where we include smart screen component and not inside the smart screen component.

krisboit (approved):

georgeburduhos (approved):

PR #1311, opened 2020-11-13
Claim all

NovacAndrei:


src/redux/ui/screens/data/actions.ts
```diff
@@ -60,7 +60,11 @@ export const fetchScreenData = (context: IScreenContext) => async (
         user: {
             os: Platform.OS as 'ios' | 'android' | 'web',
             deviceId: state.preferences.deviceId,
-            appVersion: DeviceInfo.getVersion(),
+            /**
+             * 1.4.17 - modules with promotion
+             * 1.4.18 - claim all
+             */
+            appVersion: __DEV__ ? '1.4.18' : DeviceInfo.getVersion(),
```
krisboit, line 67: there is no need for this, we can just change the version in package json as we need

krisboit (changes requested):

krisboit (approved):

georgeburduhos (approved):

PR #1322, opened 2020-11-26
Solana stake and unstake

georgeburduhos:


src/redux/wallets/reducer.ts
```diff
@@ -38,7 +38,8 @@ const newBalance = (oldBalance: any, action: any) => ({
         : new BigNumber(oldBalance?.available).toFixed(),
     total: action.data.balance
         ? new BigNumber(action.data.balance.total).toFixed()
-        : new BigNumber(oldBalance?.total).toFixed()
+        : new BigNumber(oldBalance?.total).toFixed(),
+    detailed: action.data.balance ? action.data.balance.detailed || {} : oldBalance?.detailed || {}
```
NovacAndrei, line 42: Safer to use:
`action.data.balance?.detailed || oldBalance?.detailed || {}`

src/redux/wallets/state.ts
```diff
@@ -62,6 +62,7 @@ export interface ITokenState {
         error: any;
         available: string;
         total: string;
+        detailed: any;
```
NovacAndrei, line 65: Is it possible to add a type here?

src/screens/pos-actions/pos-basic-action/pos-basic-action.tsx
```diff
@@ -131,11 +131,12 @@ export class PosBasicActionComponent extends React.Component<
 
         const tokenConfig = getTokenConfig(this.props.blockchain, this.props.token.symbol);
 
+        const balance = new BigNumber(this.props.validators[0].amountDelegated.active).plus(
```
NovacAndrei, line 134: Is this safe for all blockchains?

src/redux/wallets/reducer.ts
NovacAndrei, line 42: Safer to use:
`action.data.balance?.detailed || oldBalance?.detailed || {}`

src/redux/wallets/state.ts
NovacAndrei, line 65: Is is possible to add here a type?

Check `export interface IBlockchainTransaction<IAdditionalInfoType = any> {}`, maybe we can use something like this. To have easier tracking of detailed field.

NovacAndrei (approved):

src/redux/wallets/reducer.ts
georgeburduhos: Fixed :)

src/screens/pos-actions/pos-basic-action/pos-basic-action.tsx
georgeburduhos: yes, pending is mandatory field added when i implemented Celo, and for near and zil its returned 0

src/redux/wallets/state.ts
krisboit: It would be nice, but a bit hard in this context... it could just complicate things :)

krisboit (approved):

PR #1349, opened 2021-01-04
Total balance - cumulative

NovacAndrei:


src/redux/utils/migrations/007.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Add Cumulative Balance - 4 January 2021
+ */
+
+import { IReduxState } from '../../state';
+
+export default (state: IReduxState) => {
+    state.preferences.cumulativeBalance = false;
```
georgeburduhos, line 8: Why? there is no need for this.
If is not set - just return false

src/redux/utils/migrations/007.ts
```diff
@@ -0,0 +1,13 @@
...
+
```
georgeburduhos, line 9: Maybe use a selector, even if its a simple boolean

georgeburduhos (approved):

krisboit (approved):

PR #1389, opened 2021-02-22
min balance

georgeburduhos:


NovacAndrei (approved):

src/core/utils/api-client/validators.ts
```diff
@@ -72,13 +72,15 @@ export class ValidatorsApiClient {
         address: string,
         blockchain: Blockchain,
         chainId: string,
+        appVersion?: string,
```
NovacAndrei, line 75: Make sure on all blockchains this has been correctly.  Because `validatorId` is also optional, you need to be sure of that. Check on `Near`

NovacAndrei: image

krisboit (approved):

PR #1402, opened 2021-03-15
[WIP]Swap

georgeburduhos:


src/redux/ui/screens/data/handle-cta.ts
```diff
@@ -809,6 +812,27 @@ const handleCtaAction = async (
                     break;
                 }
 
+                case 'swap': {
```
krisboit, line 815: please don't add other `case` statements here :), use the new model (check default)

src/redux/ui/screens/data/actions/transactions/index.ts
```diff
@@ -0,0 +1,63 @@
...
+import {
+    openProcessTransactions,
+    setProcessTransactions
+} from '../../../../process-transactions/actions';
+import { buildContractCallTransaction } from './transaction-builder';
+
+export * from './transaction-builder';
+
+export interface IContractCallParams {
+    rawData: any;
```
krisboit, line 18: rename this to `params` or `args`

src/redux/ui/screens/data/actions/transactions/index.ts
```diff
@@ -0,0 +1,63 @@
...
+    blockNumber: number;
```
krisboit, line 19: why do you need this here, you can get the block number inside the builder, this way we would keep the interface cleaner

src/redux/ui/screens/data/actions/transactions/index.ts
```diff
@@ -0,0 +1,63 @@
...
+    tokenSymbol: string;
```
krisboit, line 20: I think this one can also be omitted, usually we know in what token the fees are paid

src/redux/ui/screens/data/actions/transactions/transaction-builder.ts
```diff
@@ -0,0 +1,115 @@
...
+            address: account.address,
+            publicKey: account.publicKey,
+            toAddress: contractAddress,
+            amount: params.amount,
+            feeOptions,
+            broadcastedOnBlock: params.blockNumber,
+            nonce: nonce + currentTransactions.length,
+            status: TransactionStatus.PENDING,
+            data: {
+                method: params.contractType + ' ' + tokenConfig.symbol, // TODO change to somehting relevant
```
krisboit, line 106: method should be the name of the method that was called

src/redux/ui/screens/data/actions/transactions/transaction-builder.ts
```diff
@@ -0,0 +1,115 @@
...
+                params: [contractAddress, params.amount],
```
krisboit, line 107: this should contain the list of args that were used when calling the method

georgeburduhos: we use this in transaction details, first index is the address where the call is made, second is the amount. We need to change how we handle the transaction details screen if we want to adjust this

krisboit (approved):

src/screens/token/components/default-token/default-token.tsx
```diff
@@ -118,6 +119,26 @@ export class DefaultTokenScreenComponent extends React.Component<
                         >
                             {translate('App.labels.receive')}
                         </Button>
+                        {/* TODO: maybe find a better way to handle this */}
+                        {this.props.account.blockchain === Blockchain.ZILLIQA &&
+                            token.symbol.toLowerCase() === 'gzil' && (
```
NovacAndrei, line 124: If we want to enable other swap tokens, this will be a blocker. 
Maybe is better to add here a list of supported tokens for swap by using firebase remote config.

NovacAndrei (approved):

PR #1415, opened 2021-04-07
Swap new

georgeburduhos:


src/components/widgets/components/timer-interval-price-update/timer-interval-price-update.tsx
```diff
@@ -0,0 +1,142 @@
...
+                case 'GET':
+                    response = await this.httpClient.get('');
+                    break;
+
+                default:
+                    break;
+            }
+
+            if (response?.result?.data) {
+                // response price
```
krisboit, line 88: Wll, we should do something with the prices, right ?

PR #1417, opened 2021-04-09
NEAR update tx status

NovacAndrei:


src/redux/wallets/actions/util-actions.ts
```diff
@@ -122,11 +122,12 @@ export const signAndSendTransactions = (specificIndex?: number) => async (
...
                         const txStatus = await client.utils.getTransactionStatus(outTx.id, {
                             broadcastedOnBlock: outTx.broadcastedOnBlock,
-                            currentBlockNumber
+                            currentBlockNumber,
+                            address: account.address
                         });
 
                         // if the status oif the tx is DROPPED we can reuse the nonce
-                        if (txStatus === TransactionStatus.DROPPED) {
+                        if (txStatus === TransactionStatus.DROPPED || !txStatus) {
```
NovacAndrei, line 130: @krisboit please confirm that this won't affect ZIL flow

krisboit: `getTransactionStatus` should always return a value

NovacAndrei: check again

PR #1428, opened 2021-04-19
ZIL Transaction Details Fee

NovacAndrei:


src/core/blockchain/zilliqa/client.ts
```diff
@@ -35,6 +36,27 @@ export class Client extends BlockchainGenericClient {
         this.contracts[Contracts.STAKING] = new Staking(this);
     }
 
+    public async getTransactionFees(txHash: string): Promise<IFeeOptions> {
+        try {
+            const txRes = await this.call('GetTransaction', [txHash]);
+
+            if (txRes?.result?.gasPrice && txRes?.result?.receipt?.cumulative_gas) {
+                const gasPrice = new BigNumber(txRes.result.gasPrice);
+                const gasLimit = new BigNumber(txRes.result?.receipt.cumulative_gas);
```
krisboit, line 45: well, this is not correct.
`gasLimit` is the limit it was set on the tx. `cumulative_gas` is the actual gas usage, we need them both. 
So we will have:
- gasPrice
- gasLimit
- gasUsed
- feeTotal

krisboit (changes requested):

krisboit (approved):

PR #1435, opened 2021-04-27
Send transaction log error in sentry

NovacAndrei:


georgeburduhos (approved):

src/core/blockchain/solana/client.ts
```diff
@@ -68,14 +69,20 @@ export class Client extends BlockchainGenericClient {
...
+            .then(res => {
+                if (res) {
+                    return {
+                        txHash: res,
+                        rawResponse: res
+                    };
+                }
+            })
+            .catch((error: any) => {
+                SentryCaptureException(new Error(JSON.stringify(error)));
```
krisboit, line 83: It would be nice to have in sentry error logs from all transaction broadcasts from all blockchains. So I would move this in util-actions.ts in `signAndSendTransactions()` on the `catch` clause of `sendTransaction`

NovacAndrei: It seems that we already handle errors and send them to sentry. The problem actually was that in client we don't handle errors, we don't throw them. I have fix this. Check again

PR #1456, opened 2021-05-20
Firebase & Dynamic Links

NovacAndrei:


src/core/messaging/notifications/notifications.ts
```diff
@@ -1,142 +1,146 @@
...
+
+        this.wasOpenedByNotification();
     }
 
-    public async checkPermission() {
-        // @ts-ignore seems like this is a must for now
-        const enabled = await firebase.messaging().hasPermission();
-        this.requestPermission();
+    public async getToken() {
+        return this.token || messaging().getToken();
```
georgeburduhos, line 35: this.token is string - getToken() doesnt return a Promise?

georgeburduhos (changes requested):

index.js
```diff
@@ -51,4 +52,17 @@ if (Platform.OS === 'android') {
         () => androidBgMessagingHandler
```
krisboit, line 52: since you removed the import, i think you have to remove the function call also.

src/core/messaging/notifications/notifications.ts
krisboit: yes, but the function is `async` so its result will always be a promise. You can return a `Promise<string>` or directly a `string` both will be fine.

krisboit (approved):

krisboit (approved):

krisboit (approved):

PR #1458, opened 2021-05-21
eth-fix

georgeburduhos:


src/screens/send/components/fee-options/fee-options.tsx
```diff
@@ -92,10 +99,17 @@ export class FeeOptionsComponent extends React.Component<
                 tokenSendingToken.type
             );
 
-            this.setState({ feeOptions: fees, isLoading: false });
+            this.setState({
+                feeOptions: fees,
+                isLoading: false,
+                showRetrySection: (fees.responseHasDefaults && fees.responseHasDefaults) || false
```
krisboit, line 105: this is confusing :)) I think you just need to make sure `responseHasDefaults` is a bool value. You can use: `!!fees.responseHasDefaults`

NovacAndrei (approved):

